PR #1, opened 2016-01-07
Fix typo in const SIZES

KristofferC:


KristofferC:
Nvm, I misstook what these did.

jrevels:
Hmm those are really confusing names though, come to think of it.

KristofferC:
Maybe `ITERS_LARGE_N` and `ITERS_SMALL_N` or something?

PR #7, opened 2016-02-28
add tuple sum benchmarks

KristofferC:
Benchmarks perfomance regression in https://github.com/JuliaLang/julia/issues/15277 / https://github.com/JuliaLang/julia/issues/15274


jrevels:
Thanks!

PR #14, opened 2016-06-09
edit recipe in README

jrevels:
@timholy Sorry, I meant to mention these edits in the PR review for #12, but I was too busy at the time and the PR ended up slipping my mind.


timholy:
Thanks for improving this!

PR #16, opened 2016-07-11
Ensure that ArrayLS and ArrayLSLS really are LinearSlow

timholy:
Our inlining and codegen are getting so dang awesome on julia 0.5 that recently LLVM has begun to notice that it can skip the `div` when performing linear indexing `ArrayLS` and `ArrayLSLS`. Since testing that is not really the intention of these tests, I modified the tests to defeat it. Ref https://github.com/JuliaLang/julia/pull/17355#issuecomment-231748251.

### DIFF
--- a/src/array/sumindex.jl
+++ b/src/array/sumindex.jl
@@ -147,8 +147,6 @@ end
 
 ArrayStrides1(A::Array) = ArrayStrides1(A, size(A,1))
 
-Base.similar{T}(A::ArrayLSLS, ::Type{T}, dims::Tuple{Vararg{Int}}) = ArrayLSLS(similar(A.data, T, dims))
-
 @inline Base.setindex!(A::ArrayLSLS, v, I::Int...) = A.data[I...] = v
 
 @inline Base.unsafe_setindex!(A::ArrayLSLS, v, I::Int...) = Base.unsafe_setindex!(A.data, v, I...)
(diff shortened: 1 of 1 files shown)


src/array/sumindex.jl
```diff
@@ -181,8 +186,10 @@ end
 
 function makearrays{T}(::Type{T}, r::Integer, c::Integer)
     A = samerand(T, r, c)
-    AS = ArrayLS(A)
-    ASS = ArrayLSLS(A)
+    B = similar(A, r+1, c+1)
+    B[1:r, 1:c] = A
+    AS = ArrayLS(B)
+    ASS = ArrayLSLS(B)
```
jrevels: :hushed: :smile:

timholy:
The removal of UTF16String from base seems to have broken JLD (I see what you mean...), so this won't pass on 0.5.

src/array/sumindex.jl
timholy: Hmm, maybe we should change those names :smile:

jrevels: Yeah, I didn't notice it before...maybe it's now an immortal piece of Julia benchmarking history? I'll leave it up to you to decide. 😁

timholy:
Since JLD was tagged and all of this passes, I'm going to merge (I want to test a PR :wink:).

timholy:
Out of curiosity, did https://github.com/JuliaCI/BaseBenchmarkReports/blob/9dcaa7b11fc1ab036c06ef34d488d55838e31d40/019ba50_vs_f881efb/report.md run with this, or without it? It's **exactly** the pattern I would have predicted if this was not included.

jrevels:
Ah, sorry. [Execution parameters have to be retuned before new benchmarks (or changes) land in the `nanosoldier` branch.](https://github.com/JuliaCI/BaseBenchmarks.jl#which-version-of-basebenchmarks-is-being-used-in-ci) I was going to run it tomorrow, but I can start the [retuning process](https://github.com/JuliaCI/BaseBenchmarks.jl/blob/master/etc/tune.jl) now instead. It usually takes a couple of hours.

timholy:
No hurry, I've checked locally and there doesn't seem to be a problem with that PR.

Thanks for explaining, it's nice to know the inner mysteries of the universe :8ball:

PR #20, opened 2016-08-08
fix potential nonsymmetry of cholfact benchmark input

### LINKED ISSUE #19: Running `cholfact` benchmark leads to `ArgumentError`
Hi everyone,

I was trying to run BaseBenchmarks with Julia v0.6 built from the latest commit (ec15da730a5b25431d3cfcdc04e3451a9f3e400e) on branch master and got the following the following error when running running the `cholfact` benchmark:

```
ERROR: ArgumentError: matrix is not symmetric/Hermitian. This error can be avoided by calling cholfact(Hermitian(A)) which will ignore either the upper or lower triangle of the matrix.
```

The code I used to invoke the benchmark was:

```
using BaseBenchmarks
BaseBenchmarks.load!("linalg")
run(BaseBenchmarks.SUITE[["linalg", "factorization", ("cholfact","Matrix",256)]], verbose=true);
```

I'm also using the latest master commit (e7c01ab79de57e05507978c5c69a3cea0a4b58e9) of the BaseBenchmarks repository, so I hope this is not a fault that is caused by my local setup.

Best,
Matthias

jrevels:
Thanks for finding this @MatthiasJReisinger! Since I couldn't reproduce this problem locally, would you mind checking to see if this fixes the issue on your end?

### DIFF
--- a/src/linalg/LinAlgBenchmarks.jl
+++ b/src/linalg/LinAlgBenchmarks.jl
@@ -105,7 +105,7 @@ end
 for s in SIZES
     mstr = typename(Matrix)
     m = randmat(s)
-    arr = randmat(s)'*randmat(s)
+    arr = m' * m
     g["chol", mstr, s]      = @benchmarkable chol($arr)
     g["cholfact", mstr, s]  = @benchmarkable cholfact($arr)
     g["schur", mstr, s]     = @benchmarkable schur($m)


jrevels:
I'm going to just merge this since tests passed and it can't do any harm.

MatthiasJReisinger:
It works now, thank you for fixing this!

PR #28, opened 2016-10-04
Update Nullable benchmarks to reflect isnull -> hasvalue change

nalimilan:
In current Julia master, this benchmark is slow since it needs
to compute !hasvalue. Better reflect the new structure of Nullable.
Also avoid using the two-argument Nullable() constructor so that
the code works on all Julia versions.


jrevels:
Thanks!

PR #29, opened 2016-10-04
serialization benchmarks

stevengj:
Adds benchmarks of `serialize` and `deserialize` (JuliaLang/julia#7893 and JuliaLang/julia#18633).


jrevels:
Thanks!

PR #30, opened 2016-10-06
Broadcast benchmarks

stevengj:
This adds a new suite (`broadcast`) of benchmarks testing the new broadcast-fusion functionality, the upcoming dot-operator=broadcast change, and broadcast on sparse vectors, for JuliaLang/julia#16285.

### DIFF
--- a/src/broadcast/BroadcastBenchmarks.jl
+++ b/src/broadcast/BroadcastBenchmarks.jl
@@ -52,9 +52,9 @@ g = addgroup!(SUITE, "dotop", ["broadcast!", "array"])
 perf_op_bcast!(r, x) = @dotcompat r .= 3 .* x .- 4 .* x.^2 .+ x .* x .- x .^ 3
 perf_op_bcast!(R, x, y) = @dotcompat R .= 3 .* x .- 4 .* y.^2 .+ x .* y .- x .^ 3
 
-g["dotop", "Float64", size(r), 1] = @benchmarkable perf_op_bcast!($r, $z)
-g["dotop", "Float64", size(R), 2] = @benchmarkable perf_op_bcast!($R, $x, $y)
-g["dotop", "Float64", size(r), 2] = @benchmarkable perf_op_bcast!($r, $z, 17.3)
+g["Float64", size(r), 1] = @benchmarkable perf_op_bcast!($r, $z)
+g["Float64", size(R), 2] = @benchmarkable perf_op_bcast!($R, $x, $y)
+g["Float64", size(r), 2] = @benchmarkable perf_op_bcast!($r, $z, 17.3)
 
 ###########################################################################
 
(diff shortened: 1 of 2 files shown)


stevengj:
@jrevels, it seems like there is a `.jld` file I'm supposed to update somehow when I add a new benchmark directory.  This isn't documented in the README; what am I supposed to do?

jrevels:
It's not necessary for you to update the JLD file. The failure is due to JuliaCI/BenchmarkTools.jl#23. That PR shouldn't have affected BaseBenchmarks yet, since I haven't tagged it in a release, but the [Travis script in this repo clones BenchmarkTools directly](https://github.com/JuliaCI/BaseBenchmarks.jl/blob/master/.travis.yml#L13) instead of using the latest stable release (a holdover from before BenchmarkTools was registered). I'll open a PR to fix the Travis script, and rerun the tests here.

jrevels:
The [Travis fix is merged](https://github.com/JuliaCI/BaseBenchmarks.jl/pull/31), if you'd like to rebase and rerun the tests here.

stevengj:
Okay, thanks.

src/broadcast/BroadcastBenchmarks.jl
```diff
@@ -0,0 +1,80 @@
...
+perf_bcast!(R, x, y, z) = @dotcompat R .= @compat f.(X, f.(x, y))
+
+x = randvec(10^3)
+y = randvec(10^3)'
+z = randvec(10^6)
+X = randmat(10^3)
+R = Array(Float64, length(x),length(y))
+r = similar(z)
+
+g["fusion", "Float64", size(r), 1] = @benchmarkable perf_bcast!($r, $z)
```
jrevels: As it is, the fully qualified ID for this benchmark will be ["broadcast", "fusion", ("fusion", "Float64", size(R), 1)]. Can we remove the extra fusion (same goes for the other benchmarks where the group name is duplicated in the local key)?

stevengj:
Tests are green; any other comments?

src/broadcast/BroadcastBenchmarks.jl
```diff
@@ -0,0 +1,80 @@
...
+g["Float64", size(r), 2] = @benchmarkable perf_bcast!($r, $z, 17.3)
+
+###########################################################################
+
+g = addgroup!(SUITE, "dotop", ["broadcast!", "array"])
+
+perf_op_bcast!(r, x) = @dotcompat r .= 3 .* x .- 4 .* x.^2 .+ x .* x .- x .^ 3
+perf_op_bcast!(R, x, y) = @dotcompat R .= 3 .* x .- 4 .* y.^2 .+ x .* y .- x .^ 3
+
+g["dotop", "Float64", size(r), 1] = @benchmarkable perf_op_bcast!($r, $z)
```
jrevels: `"dotop"` is redundant here for the reasons previously mentioned.

BTW, apologies for the slow review. I've been hiking on vacation, but just got back, so things should move a bit more quickly now.

jrevels: Unless I'm mistaken, it's already in a different group (the `"dotop"` group as defined on line 50). The difference in the fully qualified IDs is thus the group name.

stevengj: whoops, right, I missed that `addgroup` line

PR #32, opened 2016-10-18
add convert benchmarks

stevengj:
Add `convert` benchmarks based on JuliaLang/julia#18954

### DIFF
--- a/src/array/ArrayBenchmarks.jl
+++ b/src/array/ArrayBenchmarks.jl
@@ -211,9 +211,9 @@ x_float = 1.0 * x_int
 x_complex = x_float .+ 0.0im
 
 g = addgroup!(SUITE, "convert", ["Int"])
-g["Int", "Float64"] = @benchmarkable  perf_convert!(x_int, x_float)
-g["Float64", "Int"] = @benchmarkable  perf_convert!(x_float, x_int)
-g["Complex{Float64}", "Int"] = @benchmarkable  perf_convert!(x_complex, x_int)
-g["Int", "Complex{Float64}"] = @benchmarkable  perf_convert!(x_int, x_complex)
+g["Int", "Float64"] = @benchmarkable  perf_convert!($x_int, $x_float)
+g["Float64", "Int"] = @benchmarkable  perf_convert!($x_float, $x_int)
+g["Complex{Float64}", "Int"] = @benchmarkable  perf_convert!($x_complex, $x_int)
+g["Int", "Complex{Float64}"] = @benchmarkable  perf_convert!($x_int, $x_complex)
 
 end # module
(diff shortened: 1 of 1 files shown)


jrevels:
Looks like Travis is failing to build Blosc due to a cert error, but only when testing v0.6 (happened in #33 as well)? Any idea what's going on there?

stevengj:
Probably safe to merge this and #33, since the 0.6 failure is due to JuliaLang/julia#18997.

src/array/ArrayBenchmarks.jl
```diff
@@ -194,4 +194,26 @@ g["boolarray_true_load!"] = @benchmarkable perf_true_load!($boolarr)
...
+####################################
+
+function perf_convert!(a, x)
+    for i=1:length(x)
+        a[i] = x[i]
+    end
+    return a
+end
+
+x_int = rand(1:1000000,2000,2000)
```
jrevels: Can we make this 100 by 100 instead 2000 by 2000 (thus significantly increasing the number of samples taken within the time limit)?

stevengj:
Decreasing the benchmark size shouldn't actually improve the statistics for the mean, it will just increase the variance, right?  i.e. if it is the choice between benchmarking an operation 10x on a size-10000 array or 1000x on a size-100 array, in either case you have executed the underlying loop body the same number of times.  However, by doing the timing with finer granularity, you will get a larger variance.

stevengj:
Or does nanosoldier use the median rather than the mean?

jrevels:
Nanosoldier uses the minimum (a lot more detail on can be found in [this paper](https://arxiv.org/abs/1608.04295)). Furthermore, once I get time to fully implement [hypothesis testing in BenchmarkTools](https://github.com/JuliaCI/BenchmarkTools.jl/blob/jr/subsample/src/resample.jl), it will be important that the sample sizes are large enough to characterize an underlying timing distribution.

A less rigorous reason: it seems, from my experience, that benchmarks that touch/use less memory are less prone to the [mysterious "erroneous" timing offsets we've seen on Nanosoldier in practice](https://github.com/JuliaLang/julia/pull/18929#issuecomment-254201342) (which I believe are memory-related in nature, though I haven't pinned the exact cause down yet).

stevengj:
That's good.  (FFTW's [benchmarks](https://github.com/FFTW/fftw3/blob/fee0f966b2d3fae18019dd03a9bae338b4108d42/libbench2/speed.c) have always used the minimum as well, for much the same reason, but we never formally studied the statistics.)

stevengj:
See also [here on FFTW's benchmarking methodology](http://fftw.org/speed/method.html).

stevengj:
[LMbench](http://www.bitmover.com/lmbench/lmbench-usenix.pdf) also uses a minimum to avoid noise.  There is probably some other literature as well that you and @jiahao should take a look at for any future version of your work, although probably you guys did the most detailed statistical study.

vchuravy:
Maybe worth having scalar version of these? I was originally thinking that these would be to fast, but since https://github.com/JuliaCI/BenchmarkTools.jl/pull/23 is merged that shouldn't be a problem anymore.

stevengj:
A scalar version seems somewhat redundant; it seems like we just need one or the other.

PR #33, opened 2016-10-18
add miscellaneous benchmarks from various issues

stevengj:
This adds benchmarks based on JuliaLang/julia#13359 (cc @timholy), JuliaLang/julia#13724, JuliaLang/julia#15553, JuliaLang/julia#18135, and JuliaLang/julia#18000.

I also created a "miscellaneous" benchmark suite to stick these into.  The code can always be refactored later into different categories, but it is good to have a place to quickly add new benchmarks that don't clearly fit into one of the existing suites.


src/misc/MiscellaneousBenchmarks.jl
```diff
@@ -0,0 +1,93 @@
...
+end
+function perf_splatting(A, n, xs...)
+    s = zero(eltype(A))
+    for i = 1:n
+        s += perf_splatting_(A, xs...)
+    end
+    return s
+end
+
+g[(3,3,3)] = @benchmarkable perf_splatting($(samerand(3,3,3)), 1000000, 1, 2, 3)
```
jrevels: Can we decrease the `100000` here to `100` for the sake of efficiency (i.e. more samples per second)?

stevengj:
Okay, reduced the sample sizes.

src/misc/MiscellaneousBenchmarks.jl
```diff
@@ -0,0 +1,93 @@
...
+function perf_parse{T}(result::AbstractVector{T}, strings::AbstractVector)
+    assert(length(result) == length(strings))
+    for i = 1:length(strings)
+        @inbounds result[i] = parse(T, strings[i])
+    end
+    return result
+end
+
+g = addgroup!(SUITE, "parse", ["DateTime"])
+datestr = map(string,range(DateTime("2016-02-19T12:34:56"),Dates.Millisecond(123),10_000))
```
jrevels: This one should be dialed down as well (from `10_000` to `100` should be fine). If you're curious about the scale I'm using, the main idea is to keep most benchmarks from going over ~1ms/evaluation when possible (enabling us to gather on the order of hundreds of samples per benchmark).

stevengj: ok

jrevels:
Thanks, LGTM. Merge when ready.

PR #34, opened 2016-10-18
add some reduction benchmarks to array suite

stevengj:
Benchmarks several `reduce`-like functions to make sure we don't get regressions in this basic functionality.


jrevels:
LGTM. Ready to merge?

PR #35, opened 2016-10-19
Fix benchmark that needed state reset for every evaluation instead of every sample

jrevels:
Additionally, alter the test command to ensure this case is accounted for.


stevengj:
Whoops, sorry.

PR #37, opened 2016-10-31
Cumsum

jw3126:
Benchmark for `cumsum!`, useful for https://github.com/JuliaLang/julia/pull/18931


src/linalg/LinAlgBenchmarks.jl
```diff
@@ -63,8 +63,13 @@ for s in SIZES
     A = randmat(s)
     B = randmat(s)
     g["A_mul_B!", "Matrix{Float32}", "Matrix{Float64}", "Matrix{Float64}", s] = @benchmarkable A_mul_B!($C, $A, $B)
-end
 
+    for T in [Int32, Int64, Float32, Float64]
+        arr = rand(T, s)
```
jrevels: Could you change this to `samerand(T, s)`? That way, it'll always be the same value (ensuring that value dependence isn't a potential noise source when testing for performance regressions).

jw3126: Sure!

PR #40, opened 2016-11-08
add 6d indexing [WIP]

JaredCrean2:
This is my first stab at adding a performance test for JuliaLang/julia#18774, using a 6 dimensional array.  I have a script that auto generates these functions, so I can create as high dimension cases as you want.  I haven't created a benchmark before, so let me know if I need to make any changes.

### DIFF
--- a/src/array/ArrayBenchmarks.jl
+++ b/src/array/ArrayBenchmarks.jl
@@ -76,10 +76,7 @@ include("generate_kernel.jl")
 
 nmax = 6  # maximum dimensionality is nmax + 1
 # get path to current directory and append file name to it
-file_rev = reverse(@__FILE__)
-idx = findfirst(file_rev, '/')
-dirpath = reverse(file_rev[idx:end])
-fname = dirpath*"hdindexing.jl"
+fname = joinpath(dirname(@__FILE__), "hdindexing.jl")
 make_stencil(fname, nmax)  # generate source file
 include("hdindexing.jl")
 
--- a/src/array/ArrayBenchmarks.jl
+++ b/src/array/ArrayBenchmarks.jl
@@ -69,6 +69,33 @@ for b in values(g)
     b.params.time_tolerance = 0.50
 end
 
+
+# #18774 #
+#--------#
+include("generate_kernel.jl")
+
+nmax = 6  # maximum dimensionality is nmax + 1
+fname = "hdindexing.jl"
+make_stencil(fname, nmax)  # generate source file
+include("hdindexing.jl")
+
+npts_dir = [10000, 80, 20, 12, 9, 6]  # number of points in each direction
+for i=1:nmax
+  dims_vec = zeros(Int, i+1)
+  fill!(dims_vec, npts_dir[i])
+  dims_vec[end] = 2  # last dimension must be 2
+  u_i = samerand(dims_vec...)
+  u_ip1 = zeros(dims_vec...)
+
+  str = string(i+1, "d")
+
+  # perf_hdindexing is defined in the generated source file hdindexing.jl
+  g[str] = @benchmarkable perf_hdindexing5($u_i, $u_ip1)
+end
+
+
+
+
 # #10301 #
 #--------#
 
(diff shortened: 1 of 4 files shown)


vchuravy:
If you can automatically create these functions, maybe it might be worthwhile to put it into a macro?

src/array/ArrayBenchmarks.jl
```diff
@@ -235,4 +235,19 @@ g["Float64", "Int"] = @benchmarkable  perf_convert!($x_float, $x_int)
...
 g["Int", "Complex{Float64}"] = @benchmarkable  perf_convert!($x_int, $x_complex)
 
+
+##########################################
+# High dimensional array indexing #18774 #
+##########################################
+include("indexing.jl")
+
+g = addgroup!(SUITE, "High Dimensional Indexing", ["6d"])
+dims = (10, 10, 10, 10, 10, 2)  # trailing dimension must be 2
```
jrevels: Can we change this to something smaller, say `2` for every dimension? Discussion explaining why can be found [here](https://github.com/JuliaCI/BaseBenchmarks.jl/pull/32#issuecomment-254615441), but I should probably also put this advice in the README for future contributors.

src/array/ArrayBenchmarks.jl
```diff
@@ -235,4 +235,19 @@ g["Float64", "Int"] = @benchmarkable  perf_convert!($x_float, $x_int)
...
+u_i = rand(dims...)
```
jrevels: `rand` --> `samerand` so that the values here are deterministic (shouldn't really matter for this, but it's one less source of variance to worry about in general).

src/array/ArrayBenchmarks.jl
```diff
@@ -235,4 +235,19 @@ g["Float64", "Int"] = @benchmarkable  perf_convert!($x_float, $x_int)
 g["Complex{Float64}", "Int"] = @benchmarkable  perf_convert!($x_complex, $x_int)
 g["Int", "Complex{Float64}"] = @benchmarkable  perf_convert!($x_int, $x_complex)
 
+
+##########################################
+# High dimensional array indexing #18774 #
+##########################################
+include("indexing.jl")
+
+g = addgroup!(SUITE, "High Dimensional Indexing", ["6d"])
```
jrevels: Instead of making a new group for this, can you just add it to [the pre-existing `"index"` group](https://github.com/JuliaCI/BaseBenchmarks.jl/blob/master/src/array/ArrayBenchmarks.jl#L48)?

JaredCrean2:
I added a way to auto-generate `hdindexing.jl`, so now we can test arbitrary dimension arrays.  The generator script was set up to write to file rather than as a macro, so the code generates the file and then includes it.  

I believe I addressed all the comments.  I specified the number of points in each dimension for the 2 through 7 dimension arrays such that a run takes approximately 1/1000 of a second.  The minimum allowable dimension is 5, because it uses a 5 point finite difference stencil.

I left the maximum dimension at 7, but I think indexing should remain efficient for arrays up to MAX_TUPLETYPE_LEN - 2?

src/array/ArrayBenchmarks.jl
```diff
@@ -69,6 +69,33 @@ for b in values(g)
...
 
+
+# #18774 #
+#--------#
+include("generate_kernel.jl")
+
+nmax = 6  # maximum dimensionality is nmax + 1
+fname = "hdindexing.jl"
+make_stencil(fname, nmax)  # generate source file
+include("hdindexing.jl")
```
jrevels: This file no longer exists.

src/array/generate_kernel.jl
```diff
@@ -0,0 +1,286 @@
...
+    str *= kernel_assembly*"\n"
+  end
+
+  return str
+
+end
+
+
+function getStencil(maxdim::Integer, stencil, dim::Integer, src_eq::Integer, 
+                    second_line_indent::ASCIIString="")
```
jrevels: Can you change `ASCIIString` to `String`?

src/array/ArrayBenchmarks.jl
JaredCrean2: It is generated by `make_stencil`

jrevels: Ah. Do you know where the Travis failure is coming from then? Perhaps the `include` executes before `make_stencil`?

JaredCrean2: I'll look into it.  I suspect what is happening is the file gets written to the pwd of the shell that launched julia, rather than the directory `ArrayBenchmarks.jl` is in.

JaredCrean2:
That got it.  Good to merge?

src/array/ArrayBenchmarks.jl
```diff
@@ -69,6 +69,37 @@ for b in values(g)
...
+# #18774 #
+#--------#
+include("generate_kernel.jl")
+
+nmax = 6  # maximum dimensionality is nmax + 1
+# get path to current directory and append file name to it
+file_rev = reverse(@__FILE__)
+idx = findfirst(file_rev, '/')
+dirpath = reverse(file_rev[idx:end])
+fname = dirpath*"hdindexing.jl"
```
jrevels: Pretty nifty way of doing this, but might want to just use `dirname` + `joinpath`?

```julia
fname = joinpath(dirname(@__FILE__), "hdindexing.jl")
```

JaredCrean2: I was wondering if there was an easier way...

jrevels:
Thanks again!

PR #41, opened 2016-11-15
backwards compatible fix for recent xor change in Base

jrevels:
ref [today's daily build](https://github.com/JuliaLang/julia/commit/80770e955906c8cdbcfaf2ea73ed772195429aad#commitcomment-19824380) and  JuliaLang/julia@2751493abfae6ffe6b833b06c7950887a9379fbc

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
 julia 0.4
-BenchmarkTools 0.0.5
+BenchmarkTools 0.0.5 0.0.6-
 Compat 0.9.1
 JLD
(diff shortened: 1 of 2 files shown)


StefanKarpinski:
How about adding the definition to Compat.jl instead and upgrading that?

jrevels:
> How about adding the definition to Compat.jl instead and upgrading that?

Actually, I just realized that a `xor` fix was added to Compat.jl a couple of days ago (quite speedy!) so I've switched to using that.

tkelman:
Do we still need this BenchmarkTools upper bound?

jrevels:
Yes, BaseBenchmarks can't use a newer BenchmarkTools until Nanosoldier is updated to use the BenchmarkTools serialization wrapper. I should put that same bound should be on Nanosoldier as well.

PR #44, opened 2016-12-08
Add missing esc in dotcompat macro.

yuyichao:


jrevels:
Ah, I literally just pushed a fix and was about to open a PR. I confirmed this works on the current Julia master, thanks!

PR #45, opened 2016-12-13
Add benchmarks for sparse unary minus

Sacha0:
This pull request adds benchmarks for sparse unary minus as discussed in https://github.com/JuliaLang/julia/pull/19530. Apologies in advance for inevitable BaseBenchmarks.jl-newbie errors. Thanks and best!

### DIFF
--- a/src/sparse/SparseBenchmarks.jl
+++ b/src/sparse/SparseBenchmarks.jl
@@ -106,4 +106,18 @@ for b in values(g)
     b.params.time_tolerance = 0.3
 end
 
+##############
+# arithmetic #
+##############
+
+g = addgroup!(SUITE, "arithmetic")
+
+# unary minus, julialang repo issue #19503 / fix #19530
+g["unary minus", size(small_sqr)] = @benchmarkable -$small_sqr
+g["unary minus", size(large_sqr)] = @benchmarkable -$large_sqr
+
+for b in values(g)
+    b.params.time_tolerance = 0.3
+end
+
 end # module


src/sparse/SparseBenchmarks.jl
```diff
@@ -106,4 +106,18 @@ for b in values(g)
     b.params.time_tolerance = 0.3
 end
 
+####################
+# unary operations #
+####################
+
+g = addgroup!(SUITE, "unary ops")
```
jrevels: Hmm, the linear algebra benchmarks have an `"arithmetic"` group, can we replace `"unary ops"` with `"arithmetic"` to make things consistent?

src/sparse/SparseBenchmarks.jl
```diff
@@ -106,4 +106,18 @@ for b in values(g)
...
+
+# unary minus, julialang repo issue #19503 / fix #19530
+g["unary ops", "unary minus", size(small_sqr)] = @benchmarkable -$small_sqr
+g["unary ops", "unary minus", size(large_sqr)] = @benchmarkable -$large_sqr
```
jrevels: The extra group key here isn't necessary, e.g. `g["unary ops", "unary minus", size(small_sqr)]` can be changed to `g["unary minus", size(small_sqr)]`

src/sparse/SparseBenchmarks.jl
Sacha0: `arithmetic` is much better. Thanks!

src/sparse/SparseBenchmarks.jl
Sacha0: Fixed on push. Thanks!

jrevels:
Awesome, thanks again! I'll let you know once these benchmarks are available on Nanosoldier (they still need to undergo a `BenchmarkTools.tune!` on Nanosoldier's hardware before they'll be available).

Sacha0:
Cheers, thanks `@jrevels!`

PR #46, opened 2016-12-15
better vectorization of monte-carlo test

stevengj:
use the new-style "dot-vectorized" operators in the Monte-Carlo test in order to take advantage of JuliaLang/julia#17623, since presumably this will be the preferred style of vectorized code in performance-sensitive situations going forward.

Also, the "de-vectorized" version shouldn't allocate a temporary array for `randn()`, since that is how you would realistically want to write the optimized devec code.


jrevels:
Thanks for this (and for #47)!

PR #47, opened 2016-12-15
modern vectorization of nbody_vec

stevengj:
A more "modern" version of nbody_vec using dot operators and fusion (in 0.6).   Also, `sum(x .^ 2)` should really be `sumabs2(x)` in performance-sensitive code.

### DIFF
--- a/src/shootout/nbody_vec.jl
+++ b/src/shootout/nbody_vec.jl
@@ -6,6 +6,21 @@
 
 module NBodyVec
 
+# work around lack of Compat support for .= (Compat.jl issue #285)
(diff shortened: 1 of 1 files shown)


stevengj:
Thanks @jrevels.  How long will it be before nanosoldier benchmarks use the new version?

jrevels:
The tuning process is running now, so they should be ready in an hour or two. I'll post here when it's finished.

stevengj:
Thanks!

jrevels:
@Sacha0 @stevengj The benchmarks from #47, #46, and #45 are now runnable on Nanosoldier.

PR #49, opened 2016-12-17
fix deprecation in sumabs2 etc

KristofferC:


tkelman:
guess if we ever need to benchmark 0.4 again we can do so on a branch?

jrevels:
Thanks, @KristofferC!

> guess if we ever need to benchmark 0.4 again we can do so on a branch?

The only thing that would break on 0.4 here is `maximum(abs, ...)`, right? Maybe we can just wrap the definition of that one in a version check?

tkelman:
I think that signature would work on 0.4, but be a slower baseline than maxabs because function arguments don't work as well on 0.4

KristofferC:
Wrap everything in a version check?

jrevels:
Ah, it does work (I only checked the documentation...and the Travis build passed as well which should've tipped me off 😛). I'm okay with losing `maxabs`benchmarks etc. on v0.4. 

> Wrap everything in a version check?

It's up to you if you want to go for it, this LGTM as is.

KristofferC:
Sounds good to me.

PR #50, opened 2016-12-19
Add more DateTime parsing benchmarks

omus:
Additional benchmarks for DateTime parsing which were useful in the development of https://github.com/JuliaLang/julia/pull/19519


jrevels:
Thanks! I'll comment here once this change has passed through the tuning process, after which the benchmarks will be runnable on Nanosoldier.

jrevels:
These benchmarks are now available on Nanosoldier. Thanks again!

PR #54, opened 2016-12-30
More view performance tests

mbauman:
This adds a whole slew of tests, but I think it's very good to start tracking this more carefully.


mbauman:
Hm, I ran the array indexing benchmarks locally on my mb/too_many_indices branch and it completed in less than 10 minutes.  I think that refactor is required to constant-fold the range sums — I'm seeing 60ns for the test that timed-out master.  I'll see what I can do about 0.4 and 0.5 compatibility.

jrevels:
I'll start the tuning process now and let you know once this PR lands on Nanosoldier.

jrevels:
Done! These should run on Nanosoldier now.

PR #56, opened 2017-01-02
fix accessing T.name for jb/subtype

JeffBezanson:
Hopefully will allow the benchmarks to run on my branch. Should be backwards compatible.


tkelman:
I can merge this here but I can't deploy it to the nanosoldier nodes that run the benchmarks, that needs @jrevels to do

jrevels:
>  I can't deploy it to the nanosoldier nodes that run the benchmarks, that needs @jrevels to do

Nanosoldier pulls down the `nanosoldier` branch of BaseBenchmarks, so if you have commit access you can just merge/rebase `nanosoldier` with `master`. I just pushed this PR's commit to `nanosoldier`. 

The thing that's not automated is tuning the execution parameters for different benchmarks - that process needs to be manually triggered for new or altered benchmarks (if the alteration affects the execution time of the benchmark).

PR #57, opened 2017-01-04
Change Laplace benchmark to avoid Vector*Matrix

andreasnoack:


stevengj:
Wait, why does this patch break `ones(n) * [1 1]`?

andyferris:
@stevengj in https://github.com/JuliaLang/julia/pull/19670 I've implemented that vector * matrix makes no sense (ever).

Valid linear algebra multiplication involving matrix and vector would become:

* matrix * vector -> vector
* matrix * matrix -> matrix
* rowvector * matrix -> rowvector
* rowvector * vector -> scalar
* vector * rowvector -> matrix

I think this agrees with typical operations from undergraduate matrix algebra, Dirac notation, etc. (On the other hand, vector * matrix might be ambiguous with a 3-tensor: e.g. using standard Dirac notation in quantum mechanics for `|ψ⟩Ĥ`, the result would probably be interpreted as a 3-tensor not as a matrix. In the past, the vector * matrix operation was there only to fulfill the last operation on that list, but that is not necessary anymore.).

stevengj:
i.e. so we are no longer treating vector as a 1-column matrix...

andyferris:
In other words, `ones(n) * [1 1]` should now become `ones(n) * [1,1]'`, and `[1 1]` and `[1,1]'` are different.

andyferris:
> i.e. so we are no longer treating vector as a 1-column matrix...

Correct. This is correlated with the fact that we are no longer treating a 1-row matrix as what is now a  `RowVector`.

andyferris:
Yes, true, this is probably a typical distinction between Householder notation and Dirac notation. I feel the latter is a bit more precise, and given that Julia distinguishes scalars, vectors and matrices already we either need to make a 1x1 matrix behave like a scalar or else implement the `RowMatrix` / Dirac notation viewpoint (and I obviously have favored the latter).

andyferris:
(Though I think first-year undergrads could easily see a semantic difference between scalar *c*, vector **v**, transposed vector **v**ᵀ and matrix **A**).

stevengj:
(My experiences in attempting to teach using Dirac notation have convinced me that most physics students, most of whom claim to like it, don't actually understand it.   I can't count the number of times I've seen students do things like `Â|u⟩=|Âu⟩`...)

andreasnoack:
I haven't decided which behavior I prefer (we should probably have that discussion inhttps://github.com/JuliaLang/julia/issues/19670 ). The change here should do no harm and will hopefully allow the benchmarks to run.

PR #58, opened 2017-01-04
Add exp benchmarks

musm:
May need latest BenchmarkTools to get more meaningful timings on nanosoldier?

### DIFF
--- a/src/scalar/ScalarBenchmarks.jl
+++ b/src/scalar/ScalarBenchmarks.jl
@@ -141,6 +141,22 @@ g["significand","norm",    "Float32"] = @benchmarkable significand(1.7f0)
 g["significand","subnorm", "Float64"] = @benchmarkable significand(1.7e-310)
 g["significand","subnorm", "Float32"] = @benchmarkable significand(1.7f-40)
 
+g["exp","normal path, k = 2",              "Float 64"] = @benchmarkable exp(1.5)
+g["exp","fast path, k = 1",                "Float 64"] = @benchmarkable exp(0.5)
+g["exp","no agument reduction, k = 9",     "Float 64"] = @benchmarkable exp(0.1)
+g["exp","small argument path",             "Float 64"] = @benchmarkable exp(2.0^-30)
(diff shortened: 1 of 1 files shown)


src/scalar/ScalarBenchmarks.jl
```diff
@@ -145,4 +145,30 @@ for b in values(g)
...
+g["exp","overflow",                        "Float 64"] = @benchmarkable exp(900.0)
+g["exp","underflow",                       "Float 64"] = @benchmarkable exp(-900.0)
+
+g["exp","normal path, k = 2",              "Float 32"] = @benchmarkable exp(1f5)
+g["exp","fast path, k = 1",                "Float 32"] = @benchmarkable exp(0f5)
+g["exp","no agument reduction, k = 9",     "Float 32"] = @benchmarkable exp(0f1)
+g["exp","small argument path",             "Float 32"] = @benchmarkable exp(2f0^-15)
+g["exp","normal path -> small, k = -1045", "Float 32"] = @benchmarkable exp(-724f0)
+g["exp","overflow",                        "Float 32"] = @benchmarkable exp(150f0)
+g["exp","underflow",                       "Float 32"] = @benchmarkable exp(-150f0)
```
jrevels: Can we keep these in the same group as the other scalar exponentiation benchmarks? Maybe just change the current `"floatexp"` group to `"exp"`?

src/scalar/ScalarBenchmarks.jl
```diff
@@ -145,4 +145,30 @@ for b in values(g)
     b.params.time_tolerance = 0.40
 end
 
+#######
+# exp #
+#######
+
+g = addgroup!(SUITE, "exp")
+
+g["exp","normal path, k = 2",              "Float 64"] = @benchmarkable exp(1.5)
```
jrevels: You can leave out the additional `"exp"` here. It's redundant since `"exp"` is already the group ID.

musm:
k fixed thanks

jrevels:
Sorry for the delay - these benchmarks are now on Nanosoldier.

PR #61, opened 2017-01-25
RFC add matrix types, add sqrtm benchmarks

felixrehren:
Start benchmarking `sqrtm`, cf. https://github.com/JuliaLang/julia/pull/20214
Add `UnitUpperTriangular` and `NPDUpperTriangular` (NPD = non-positive-definite) matrices to the testing

### DIFF
--- a/src/linalg/LinAlgBenchmarks.jl
+++ b/src/linalg/LinAlgBenchmarks.jl
@@ -7,6 +7,7 @@ using BenchmarkTools
 using Compat
 
 import Compat: UTF8String, view
+import Base.LinAlg: UnitUpperTriangular
 
 const SUITE = BenchmarkGroup(["array"])
 
--- a/src/linalg/LinAlgBenchmarks.jl
+++ b/src/linalg/LinAlgBenchmarks.jl
@@ -85,6 +85,12 @@ for s in SIZES
         g["cumsum!", T, s] = @benchmarkable cumsum!($arr, $arr)
     end
 
+    for M in (UpperTriangular, UnitUpperTriangular, NPDUpperTriangular, Hermitian)
+        mstr = typename(M)
+        m = linalgmat(M, s)
+        g["sqrtm", mstr, s] = @benchmarkable sqrtm($m)
+    end
+
 end
 
 for b in values(g)
@@ -134,24 +140,6 @@ for s in SIZES
     g["schurfact", mstr, s] = @benchmarkable schurfact($m)
     g["qr", mstr, s]        = @benchmarkable qr($m)
     g["qrfact", mstr, s]    = @benchmarkable qrfact($m)
-end
-
-for b in values(g)
-    b.params.time_tolerance = 0.45
-end
-
-##############
-# operations #
-##############
-
-g = addgroup!(SUITE, "operations")
-
-for M in (UpperTriangular, UnitUpperTriangular, NPDUpperTriangular, Hermitian)
-    mstr = typename(M)
-    for s in SIZES
-        m = linalgmat(M, s)
-        g["sqrtm", mstr, s] = @benchmarkable sqrtm($m)
-    end
 end
 
 for b in values(g)
--- a/src/linalg/LinAlgBenchmarks.jl
+++ b/src/linalg/LinAlgBenchmarks.jl
@@ -32,6 +32,24 @@ function linalgmat(::Type{Hermitian}, s)
     return Hermitian(A + A')
 end
 
+function linalgmat{M}(::Type{M}, s, f::Function)
+    A = linalgmat(M, s)
+    for i in 1:s
+        A[i,i] = f(A[i,i])
+    end
+    return A
+end
+linalgmat{T}(::Type{T}, s, identity) = linalgmat(T, s)
+
+linalgmat(::Type{UnitUpperTriangular}, s) = linalgmat(UpperTriangular, s, x -> 1)
+
+# Non-positive-definite upper-triangular matrix
+type NPDUpperTriangular
+end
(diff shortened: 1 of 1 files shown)


src/linalg/LinAlgBenchmarks.jl
```diff
@@ -32,6 +32,24 @@ function linalgmat(::Type{Hermitian}, s)
...
 end
 
+function linalgmat{M}(::Type{M}, s, f::Function)
+    A = linalgmat(M, s)
+    for i in 1:s
+        A[i,i] = f(A[i,i])
+    end
+    return A
+end
+linalgmat{T}(::Type{T}, s, identity) = linalgmat(T, s)
```
stevengj: i don't think this will do what you want, you need `::typeof(identity)`

src/linalg/LinAlgBenchmarks.jl
```diff
@@ -32,6 +32,24 @@ function linalgmat(::Type{Hermitian}, s)
...
+
+linalgmat(::Type{UnitUpperTriangular}, s) = linalgmat(UpperTriangular, s, x -> 1)
+
+# Non-positive-definite upper-triangular matrix
+type NPDUpperTriangular
+end
+linalgmat(::Type{NPDUpperTriangular}, s) = linalgmat(UpperTriangular, s, x -> randn()*x)
```
stevengj: usually the benchmarks should use random numbers with a fixed seed, so that they are the same between benchmark runs.  see the functions in `RandUtils.jl`

src/linalg/LinAlgBenchmarks.jl
```diff
@@ -32,6 +32,24 @@ function linalgmat(::Type{Hermitian}, s)
...
+function linalgmat{M}(::Type{M}, s, f::Function)
+    A = linalgmat(M, s)
+    for i in 1:s
+        A[i,i] = f(A[i,i])
+    end
+    return A
+end
+linalgmat{T}(::Type{T}, s, identity) = linalgmat(T, s)
+
+linalgmat(::Type{UnitUpperTriangular}, s) = linalgmat(UpperTriangular, s, x -> 1)
```
stevengj: This doesn't actually return a `UnitUpperTriangular` matrix type?

felixrehren:
@stevengj Thanks a lot for your comments -- I really appreciate it!
I eliminated some of the complexity I introduced. The issue is that `sqrtm` behaves differently on non/positive-definite upper-triangular matrices, so a new non-positive-definite matrix to test is essential. I'm not sure if my implementation is good, suggestions welcome.

src/linalg/LinAlgBenchmarks.jl
```diff
@@ -124,6 +140,24 @@ for b in values(g)
     b.params.time_tolerance = 0.45
 end
 
+##############
+# operations #
+##############
+
+g = addgroup!(SUITE, "operations")
```
jrevels: Can we change `"operations"` to a more descriptive group name? The word `"operations"` could apply to basically any benchmark.

src/linalg/LinAlgBenchmarks.jl
```diff
@@ -124,6 +140,24 @@ for b in values(g)
...
+for M in (UpperTriangular, UnitUpperTriangular, NPDUpperTriangular, Hermitian)
+    mstr = typename(M)
+    for s in SIZES
+        m = linalgmat(M, s)
+        g["sqrtm", mstr, s] = @benchmarkable sqrtm($m)
+    end
+end
+
+for b in values(g)
+    b.params.time_tolerance = 0.45
```
jrevels: Do you have any reason to believe that the benchmarks will be this noisy? It's probably better to stick with the default, and then we raise the tolerance threshold later if we need to.

felixrehren: None at all, I didn't know what I was doing -- I copied them from above, but is it ok just to remove these lines?

src/linalg/LinAlgBenchmarks.jl
felixrehren: Absolutely, any suggestion? I struggled with this

jrevels: I would just put it in the `"arithmetic"` group.

src/linalg/LinAlgBenchmarks.jl
jrevels: Yes. The `time_tolerance` parameter basically tells `BenchmarkTools.judge` the percent threshold at which something is classified as a "regression" or "improvement".

felixrehren: Done. Does the same `time_tolerance = 0.45` apply in the `arithmetic` group?

jrevels: Oops, yeah, it does. The real solution I guess is to add a subgroup under `"arithmetic"` for the binary arithmetic benchmarks (`*`, `+`, `-`, etc.), which were what that time tolerance setting was meant for. I can do that in a future PR, though, let's just get this merged.

felixrehren: Sweet. Travis is good for 0.4, 0.5 but [fail on 0.6](https://travis-ci.org/JuliaCI/BaseBenchmarks.jl/jobs/198065838) seemingly due to

    ERROR: LoadError: LoadError: UndefVarError: readbytes not defined

Is it related? Could it be related to a 0.6-nightly change in the last week?

jrevels: It's not from this PR, the error is occurring in FileIO.jl. 

BaseBenchmarks and it's dependencies haven't been runnable on Julia master in a couple days now, there have just been too many changes that broke the stack (plus all the type inference bugs).

PR #63, opened 2017-02-10
bench mapreduce with array accesses

felixrehren:
Cf. https://github.com/JuliaLang/julia/issues/20517:
`mapreduce` can be much slower than the equivalent `for`-loop, currently mostly coming from an inference bug https://github.com/JuliaLang/julia/issues/15276 related to `Core.box`.

This benchmark would allow keeping an eye on this realistic code pattern. If that is what you had in mind, @stevengj ?

### DIFF
--- a/src/array/ArrayBenchmarks.jl
+++ b/src/array/ArrayBenchmarks.jl
@@ -60,7 +60,7 @@ for (A, str) in (arrays_iter..., ranges_iter...)
     g["sumlinear_view", str]     = @benchmarkable perf_sumlinear_view($A)
     g["sumcartesian_view", str]  = @benchmarkable perf_sumcartesian_view($A)
     if ndims(A) == 2
-        g["mapr_access", str]    = @benchmarkable perf_mapr_access($A) #20517
+        g["mapr_access", str]    = @benchmarkable perf_mapr_access(A, B, zz, n) setup = $setup_mapr_access #20517
     end
     if ndims(A) <= 2
         g["sumcolon", str]       = @benchmarkable perf_sumcolon($A)
--- a/src/array/ArrayBenchmarks.jl
+++ b/src/array/ArrayBenchmarks.jl
@@ -60,7 +60,7 @@ for (A, str) in (arrays_iter..., ranges_iter...)
     g["sumlinear_view", str]     = @benchmarkable perf_sumlinear_view($A)
     g["sumcartesian_view", str]  = @benchmarkable perf_sumcartesian_view($A)
     if ndims(A) == 2
-        g["mapr_access", str]    = @benchmarkable perf_mapr_access(A, B, zz, n) setup = $setup_mapr_access #20517
+        g["mapr_access", str]    = @benchmarkable perf_mapr_access(A, B, zz, n) setup = (B, zz, n = setup_mapr_access($A)) #20517
     end
     if ndims(A) <= 2
         g["sumcolon", str]       = @benchmarkable perf_sumcolon($A)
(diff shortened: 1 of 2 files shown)


src/array/sumindex.jl
```diff
@@ -191,6 +191,18 @@ function perf_ind2sub(sz, lrange)
     si, sj, sk
 end
 
+function perf_mapr_access(A)
+    z = zero(eltype(A))
+    zz = z*z
+    n = Base.LinAlg.checksquare(A)
+    B = Vector{typeof(zz)}(n)
```
jrevels: If the benchmark is testing the performance of `mapreduce`, it'd be better not to benchmark this setup step, no? See [BenchmarkTools' setup/teardown phase functionality](https://github.com/JuliaCI/BenchmarkTools.jl/blob/master/doc/manual.md#setup-and-teardown-phases).

felixrehren: Like this?

src/array/ArrayBenchmarks.jl
```diff
@@ -59,6 +59,9 @@ for (A, str) in (arrays_iter..., ranges_iter...)
     g["sumeach_view", str]       = @benchmarkable perf_sumeach_view($A)
     g["sumlinear_view", str]     = @benchmarkable perf_sumlinear_view($A)
     g["sumcartesian_view", str]  = @benchmarkable perf_sumcartesian_view($A)
+    if ndims(A) == 2
+        g["mapr_access", str]    = @benchmarkable perf_mapr_access(A, B, zz, n) setup = $setup_mapr_access #20517
```
jrevels: The reason this is failing is due to the weird scoping rules of `setup`; you'd need to interpolate `A` into the expression for it to work. It's pretty ugly, unfortunately. For example:

```julia
julia> for a in 1:3
           @benchmark sin(b) setup=(println(a); b = 1 + a) samples=1 evals=1
       end
ERROR: UndefVarError: a not defined
```

vs.

```julia
# this one should work
julia> for a in 1:3
           @benchmark sin(b) setup=(println($a); b = 1 + $a) samples=1 evals=1
       end
1
1
2
2
3
3
```

Anyway, instead of interpolating an expression here, I would just make `setup_mapr_access` a function and  write `setup = (B, zz, n = setup_mapr_access())`.

src/array/ArrayBenchmarks.jl
```diff
@@ -59,6 +59,9 @@ for (A, str) in (arrays_iter..., ranges_iter...)
     g["sumeach_view", str]       = @benchmarkable perf_sumeach_view($A)
     g["sumlinear_view", str]     = @benchmarkable perf_sumlinear_view($A)
     g["sumcartesian_view", str]  = @benchmarkable perf_sumcartesian_view($A)
+    if ndims(A) == 2
+        g["mapr_access", str]    = @benchmarkable perf_mapr_access(A, B, zz, n) setup = (B, zz, n = setup_mapr_access($A)) #20517
```
jrevels: Argh, looks like you've discovered a new BenchmarkTools bug. [I just posted an issue for it in the BenchmarkTools repo](https://github.com/JuliaCI/BenchmarkTools.jl/issues/44). Sorry about this.

You should be able to fix this by using `begin...end` instead of parens for `setup`, e.g. `setup = begin B, zz, n = setup_mapr_access($A) end`.

felixrehren: Haha, no worries for me -- thanks for your patience with this. Fix incoming

src/array/ArrayBenchmarks.jl
```diff
@@ -59,6 +59,9 @@ for (A, str) in (arrays_iter..., ranges_iter...)
     g["sumeach_view", str]       = @benchmarkable perf_sumeach_view($A)
     g["sumlinear_view", str]     = @benchmarkable perf_sumlinear_view($A)
     g["sumcartesian_view", str]  = @benchmarkable perf_sumcartesian_view($A)
+    if ndims(A) == 2
+        g["mapr_access", str]    = @benchmarkable perf_mapr_access(A, B, zz, n) setup = begin B, zz, n = setup_mapr_access($A) end #20517
```
jrevels: `A` here needs to be interpolated since it's coming from the loop scope rather than the `setup` scope.

jrevels:
Awesome, looks like tests are passing now! Let me know once you're ready to merge.

felixrehren:
LGTM but would like to make sure someone else, like @stevengj, has looked at this ?

PR #64, opened 2017-02-14
Actually run `perf_lucompletepivSub!`

martinholters:
...instead of running `perf_lucompletepivCopy!` twice.

Ref https://github.com/JuliaCI/BaseBenchmarks.jl/commit/da9fda347309e3ee47437cfa2e249171109bf79f#commitcomment-20858355


martinholters:
Looks like this uncovered a bug in `perf_lucompletepivSub!`, fix added.

jrevels:
Nice catch! Thanks for the fix.

PR #66, opened 2017-03-06
add benchmarks for broadcast including types in args, ref #20802

KristofferC:


Sacha0:
Might also be worth adding benchmarks for, e.g., `broadcast(+, 1, tup)`, `broadcast(+, tup, tup)`, and `broadcast(+, 1, tup, 1, tup, 1, tup)`? These were a few of the test cases that proved interesting in https://github.com/JuliaLang/julia/issues/20817. (Not suggesting these need to go in this PR, but if that sounds like fun! :) ). Best!

jrevels:
Does `@compat` fix the v0.5 failure for the new n-ary dot broadcast being benchmarked here? If not, you can just add a `VERSION` check.

KristofferC:
I'll add some more benchmarks and check with Compat.

KristofferC:
Added some benchmarks. DId not get Compat to work.

src/broadcast/BroadcastBenchmarks.jl
```diff
@@ -62,4 +62,32 @@ g[size(S), 2] = @benchmarkable perf_sparse_op($S, $S)
...
+
+f_round(v) = @compat round.(Int, v)
+
+g = addgroup!(SUITE, "typeargs", ["broadcast", "array"])
+r1, r2, r3 = rand(3), rand(5), rand(10)
+for r in (r1, r2, r3)
+    g[length(r)] = @benchmarkable f_round($r)
+end
+
+g = addgroup!(SUITE, "typeargs", ["broadcast", "tuple"])
```
jrevels: These groups have the same keys, so I believe the second will overwrite the first. My suggestion is to just make one `"typeargs"` group,  and put all of these `"typeargs"` benchmarks there.

src/broadcast/BroadcastBenchmarks.jl
```diff
@@ -62,4 +62,32 @@ g[size(S), 2] = @benchmarkable perf_sparse_op($S, $S)
 
 ###########################################################################
 
+if VERSION > v"0.6-"
```
jrevels: Can you make this VERSION check more specific? It seems people are doing a lot of bisect-benchmarking these days, so I fear they'd hit an inconsistency if this check didn't explicitly correspond to the necessary commit.

src/broadcast/BroadcastBenchmarks.jl
KristofferC: I think I muddled up key vs tag... Will try fix.

KristofferC: Pushed a new version, hopefully got it right this time,

src/broadcast/BroadcastBenchmarks.jl
KristofferC: Since I haven't been involved in the implementation of this, tracking down the exact commits for the two functions is difficult. I could set up some bisection script but it will be a while until I get around that I think.

Sacha0: For all of the benchmarks in this block to be fast, I think you need at least [this](https://github.com/JuliaLang/julia/commit/2e12d10097750eaab89bd5caa4a515bc5e8b92ef) or a more recent commit? Best!

KristofferC:
updated the commit number

jrevels:
Thanks!

PR #67, opened 2017-03-07
add benchmarks for fixed size arrays

KristofferC:
Add Benchmarks for a small fixed size array (FSA) implementation (based on tuples) with some reduce functionality and mat * mat/vect. 
I feel this is good to have here since we tuples are the de facto way that FSA's are implemented.

I borrowed some of your linear algebra code @andyferris. Hope you don't mind.


KristofferC:
Oh, I didn't realize this still tested against 0.4.

andyferris:
> Hope you don't mind.

Of course not! This seems like a good idea.

PR #68, opened 2017-03-26
Benchmark for backsolve of sparse hermitian matrices

ranjanan:
Ref https://github.com/JuliaLang/julia/pull/21165

### DIFF
--- a/src/sparse/SparseBenchmarks.jl
+++ b/src/sparse/SparseBenchmarks.jl
@@ -120,17 +120,4 @@ for b in values(g)
     b.params.time_tolerance = 0.3
 end
 
-#############
-# Backsolve #
-#############
-
-g = addgroup!(SUITE, "backsolve")
-
-A = sprand(10^4, 10^4, 1e-4)
-A = A + A' + speye(10^4)
-@assert ishermitian(A)
-b = rand(10^4)
-
-g["hermitiansolve"] = @benchmarkable A \ b
-
 end # module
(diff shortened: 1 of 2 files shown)


src/sparse/SparseBenchmarks.jl
```diff
@@ -120,4 +120,17 @@ for b in values(g)
...
+#############
+# Backsolve #
+#############
+
+g = addgroup!(SUITE, "backsolve")
+
+A = sprand(10^4, 10^4, 1e-4)
+A = A + A' + speye(10^4)
+@assert ishermitian(A)
+b = rand(10^4)
```
jrevels: Can you use `BaseBenchmark.RandUtils` to construct these arrays instead of calling an unseeded `rand`/`srand`?

src/sparse/SparseBenchmarks.jl
```diff
@@ -120,4 +120,17 @@ for b in values(g)
     b.params.time_tolerance = 0.3
 end
 
+#############
+# Backsolve #
+#############
+
+g = addgroup!(SUITE, "backsolve")
```
jrevels: Instead of adding a new benchmark group here, it probably makes sense to just add this to [the group where backsolve is normally tested](https://github.com/JuliaCI/BaseBenchmarks.jl/blob/master/src/linalg/LinAlgBenchmarks.jl#L55), which you can do by following a pattern similar to [here](https://github.com/JuliaCI/BaseBenchmarks.jl/blob/master/src/linalg/LinAlgBenchmarks.jl#L37).

tkelman:
@jrevels is this okay now?

jrevels:
I'll kick off the tuning process now, and comment here when it's done.

jrevels:
This benchmark should now be available via Nanosoldier.

PR #70, opened 2017-04-01
add prevpow2 and nextpow2 benchmarks

rfourquet:
WIP as I'm not familiar at all with the framework. @tkelman [asked me](https://github.com/JuliaLang/julia/pull/21231#issuecomment-290355981) if benchmarks exist for those function, here they are. I'm not sure about pretty much everything:

1. the `tol` parameter, which I copied verbatim from other benchmarks;
2. the name "intfuncs" of the group (after the base/intfuncs.jl file where the functions are defined), and the choice of the "scalar" group;
3. the name of `tstr` containing a sign (e.g. "BigInt +") (the perfs depend on the sign of the argument);
4. whether to test multiple values, to get a sort of average, instead of only one value.

### DIFF
--- a/src/scalar/ScalarBenchmarks.jl
+++ b/src/scalar/ScalarBenchmarks.jl
@@ -148,7 +148,7 @@ g["exp","small argument path",             "Float 64"] = @benchmarkable exp(2.0^
 g["exp","normal path -> small, k = -1045", "Float 64"] = @benchmarkable exp(-724.0)
 g["exp","overflow",                        "Float 64"] = @benchmarkable exp(900.0)
 g["exp","underflow",                       "Float 64"] = @benchmarkable exp(-900.0)
- 
+
 g["exp","normal path, k = 2",              "Float 32"] = @benchmarkable exp(1f5)
 g["exp","fast path, k = 1",                "Float 32"] = @benchmarkable exp(0f5)
 g["exp","no agument reduction, k = 9",     "Float 32"] = @benchmarkable exp(0f1)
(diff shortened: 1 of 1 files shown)


src/scalar/ScalarBenchmarks.jl
```diff
@@ -161,4 +161,21 @@ for b in values(g)
...
+for T in INTS
+    x = T[0, 1, 2, 3, 4, 10, 100, 1000, 10000, 100000]
+
+    tol = in(T, BIGNUMS) ? 0.40 : 0.25
+    for sgn in (+, -)
+        tstr = string(T, " ", sgn)
+        g["prevpow2", tstr] = @benchmarkable map(prevpow2, $(sgn(x))) time_tolerance=tol
+        g["nextpow2", tstr] = @benchmarkable map(nextpow2, $(sgn(x))) time_tolerance=tol
+    end
+end
```
jrevels: I would change this to the following:

```julia
for T in INTS
    x = T[0, 1, 2, 3, 4, 10, 100, 1000, 10000, 100000]
    y = similar(x)
    tol = in(T, BIGNUMS) ? 0.40 : 0.25
    tstr = string(T)
    for sgn in (+, -)
        sgnstr = string(sgn)
        g["prevpow2", tstr, sgnstr] = @benchmarkable map!(prevpow2, $y, $(sgn(x))) time_tolerance=tol
        g["nextpow2", tstr, sgnstr] = @benchmarkable map!(nextpow2, $y, $(sgn(x))) time_tolerance=tol
    end
end
```

The two important changes here are:

- splitting up the type and sign in the benchmark ID
- Using in-place `map!` instead of allocating a temporary for each benchmark execution. This should cut down on noise and makes the benchmark more targeted (i.e. it avoids benchmarking allocation overhead).

rfourquet:
Thank you for the review! I updated according to your request and added some new values to test. Also I included the function to test as a for-loop variable, which is not in the style of the file, so I can revert that bit.

jrevels:
For future reference, it's usually cleaner to manually unroll such loops when the number of combined items are so small and are known at program time. It doesn't matter enough to hold up this PR though. Thanks again for the contribution!

rfourquet:
Ok and thank you!

PR #72, opened 2017-04-10
retune benchmarks (includes #70)

jrevels:
Once this is merged into `master` (and subsequently merged into the `nanosoldier` branch), then the benchmarks from #70 will be available on Nanosoldier. 

@rfourquet Nothing was run [here](https://github.com/JuliaLang/julia/pull/21231#issuecomment-292764647) because the benchmarks hadn't been deployed to Nanosoldier yet.


rfourquet:
Ok thank you for helping!

PR #73, opened 2017-04-20
Add benchmarks for exp10 function

musm:
please squash


jrevels:
Judging from https://github.com/JuliaCI/BenchmarkTools.jl/issues/56, it seems like your intent here isn't actually to allow the expressions here to be constant-folded? You can ensure they aren't by interpolating all the constants (which tells BenchmarkTools to treat it like an argument passed into the wrapper function surrounding the benchmark expression).

musm:
I think in this case all the other scalar benchmarks should be changed in the floatexp group, in particular  exp, ldexp

I don't think we want to benchmark constant folding in this case although i'm not 100% sure.
But this test parallels the `exp` right above so it may not matter.

BTW the original issue was prompted by a different test ( I was looking at evaluating a function taking as input a vector)

jrevels:
Okay, let's merge this and I'll fix the constant-folding in a subsequent PR.

PR #77, opened 2017-06-01
Add sparse benchmarks for AbstractUnitRange

timholy:
Ref https://github.com/JuliaLang/julia/pull/22175

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,11 +1,13 @@
 environment:
   matrix:
-  - JULIAVERSION: "julialang/bin/winnt/x86/0.4/julia-0.4-latest-win32.exe"
-  - JULIAVERSION: "julialang/bin/winnt/x64/0.4/julia-0.4-latest-win64.exe"
-  - JULIAVERSION: "julialang/bin/winnt/x86/0.5/julia-0.5-latest-win32.exe"
-  - JULIAVERSION: "julialang/bin/winnt/x64/0.5/julia-0.5-latest-win64.exe"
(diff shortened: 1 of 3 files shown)


tkelman:
guess we need to make these conditional on `isdefined(Base, :OneTo)` ?

appveyor.yml
```diff
@@ -4,6 +4,8 @@ environment:
   - JULIAVERSION: "julialang/bin/winnt/x64/0.4/julia-0.4-latest-win64.exe"
   - JULIAVERSION: "julialang/bin/winnt/x86/0.5/julia-0.5-latest-win32.exe"
   - JULIAVERSION: "julialang/bin/winnt/x64/0.5/julia-0.5-latest-win64.exe"
+  - JULIAVERSION: "julialang/bin/winnt/x86/0.6/julia-0.5-latest-win32.exe"
+  - JULIAVERSION: "julialang/bin/winnt/x64/0.6/julia-0.5-latest-win64.exe"
```
Evizero: should these links be updated?

Evizero: (aside from the typo mixing 0.6 with 0.5 in one line)

timholy: Good catch

appveyor.yml
```diff
@@ -21,7 +23,7 @@ notifications:
 install:
```
Evizero: One more tiny nitpick. This line will be needed for https I think

```
 - ps: "[System.Net.ServicePointManager]::SecurityProtocol = [System.Net.SecurityProtocolType]::Tls12"
```

timholy: Thanks. Added.

PR #82, opened 2017-07-10
add RNG benchmarks

rfourquet:
Tests most of the functionalities from the Base.Random module.


jrevels:
Awesome stuff, thanks for all of these!

Looks mostly good to me after a skim. Can you make these benchmarks backwards compatible to get tests passing (note that we can drop v0.4 support if we absolutely need to)?

Also - especially for the unseeded `rand` calls - are we sure that performance of these benchmarks are deterministic/reproducible between measurements?

rfourquet:
I tried to make compatible with older versions, but I don't have them installed, so will have to wait for CI to tell me about other incompatibilities. I always used the same version number (as a lower bound) for newer functionalities, without bothering to find the exact commit introducing those, I hope this will be fine.

> Also - especially for the unseeded rand calls - are we sure that performance of these benchmarks are deterministic/reproducible between measurements?

As far as Base is concerned, this should be fine (in e.g. Primes.jl, this would be a very different story). But do you confirm that the time reported for a benchmark is the average of multiple "evals" of a single "sample"? I could set this number of "evals" to at least 5 for example, but it may be cleaner to re-seed the RNGs before each benchmark when possible. Unless the number of evals per benchmark and the order of the benchmarks are deterministic (between 2 invocations, e.g. for 2 different branches), then seeding once at the beginning of the file could be enough? Will do as you feel is appropriate.

tkelman:
It would be best to avoid having the code here not work on early prereleases of anything. Better to have slightly fewer benchmarks on those versions of julia than have it not able to run. So use known-working floors if you aren't going to find a specific version.

jrevels:
> But do you confirm that the time reported for a benchmark is the average of multiple "evals" of a single "sample"? I could set this number of "evals" to at least 5 for example, but it may be cleaner to re-seed the RNGs before each benchmark when possible. Unless the number of evals per benchmark and the order of the benchmarks are deterministic (between 2 invocations, e.g. for 2 different branches), then seeding once at the beginning of the file could be enough? Will do as you feel is appropriate.

Just to be clear on terminology: a single data point ("sample" in BenchmarkTools terms) is one timing measurement divided by the number of benchmark evaluations that occurred during the time interval. The number of evaluations per sample is set by [the pre-tuning process](https://github.com/JuliaCI/BenchmarkTools.jl/blob/master/doc/manual.md#caching-parameters) (so, cached in `BaseBenchmarks/etc/params.jld`). Actually running a benchmark collects as many samples as possible giving the time/sample budget, after which a variety of summary statistics can be calculated. The one used by Nanosoldier for benchmark comparison is `minimum`. 

As for this specific case:  Unless these benchmarks actually seek to measure the performance of executing `rand` with a specific seed, we probably shouldn't seed the benchmarks, right? This assumes that RNG performance isn't usually wildly nondeterministic given random seeds, which sounds correct to me, but I'm no expert 😛

rfourquet:
> This assumes that RNG performance isn't usually wildly nondeterministic given random seeds

Yes this is correct, so indeed I favor not seeding explicitly the RNGs (in particular as `RandomDevice` cannot be seeded). I can see two subtle points to consider, but which seem to not be real problems according to what you explained above (thanks btw!):

1) `rand(Float64)` uses a cache of 382 random values; once all the values are consumed, a new set of 382 random values are produced; this means that if in one benchmark (say PR), there was e.g. 300 evals per sample, and in another benchmark (say master) there was 400 evals per sample, then there could be differences: for PR, the best samples would not have to call the expensive routine filling up the cache (re-generating 382 random values), while on master even the best sample would always have to re-fill the cache at least once. Most rand-related methods rely on this. But given the pre-tuning procees, I assume the number of eval/sample is fixed across benchmarks.
2) for some functions (e.g. `rand(::Set)`), the performance should be stable on average, but not if there was a single eval per sample: one eval could need to use up to e.g. 10 or 20 calls to `rand(::Float)` internally, while the most lucky evals could need only one such call. But overall, these are cheap functions, so I assume there will always be a sufficiently hight number of evals/sampe to make this a non-issue (please correct me if this assumption is wrong).

rfourquet:
> So use known-working floors if you aren't going to find a specific version.

Yes this is what I did: I used a recent master (will update to today's master for being more sure) as a floor for recent new functionalities, and v"0.5-" for a functionality that was added in a commit (dbba1d47f3facc23b8d351d407553c576f9f658a for the record) which is an ancestor of tag "v0.5.0-rc0"; is this safe?

tkelman:
No, `v"0.5-"` includes the very first commit after release-0.4 branched and master became 0.5-dev

tkelman:
Since that commit was a (squashed) PR merge commit, you can run `contrib/commit-name.sh dbba1d47f3facc23b8d351d407553c576f9f658a` and get the exact version, `v"0.5.0-pre+5657"`

rfourquet:
Ok perfect, thanks!

rfourquet:
Bump

jrevels:
Really great work, thanks again for this!

rfourquet:
And thanks for your amazing framework!

PR #84, opened 2017-07-11
added some julia parser etc benchmarks

stevengj:
Inspired by JuliaLang/julia#22089


stevengj:
Travis failure is unrelated HDF5 problem?

PR #87, opened 2017-07-25
Add acos and asin benchmarks.

pkofod:


src/scalar/ScalarBenchmarks.jl
```diff
@@ -445,4 +446,45 @@ for T in (Float64, )# (Float32, Float64) add Float32 later
...
+########
+# asin #
+########
+
+g = addgroup!(SUITE, "asin")
+for T in (Float32, Float64)
+    _arg_string = arg_string(T)
+    g["zero", _arg_string] = @benchmarkable asin($(zero(T)))
+    g["small", "positive argument", _arg_string] = @benchmarkable asin($(nextfloat(zero(T))))
+    g["small", "negative argument", _arg_string] = @benchmarkable asin($(nextfloat(zero(T))))
```
Sacha0: `prevfloat`?

src/scalar/ScalarBenchmarks.jl
```diff
@@ -445,4 +446,45 @@ for T in (Float64, )# (Float32, Float64) add Float32 later
...
+########
+# acos #
+########
+
+g = addgroup!(SUITE, "acos")
+for T in (Float32, Float64)
+    _arg_string = arg_string(T)
+    g["zero", _arg_string] = @benchmarkable acos($(zero(T)))
+    g["small", "positive argument", _arg_string] = @benchmarkable acos($(nextfloat(zero(T))))
+    g["small", "negative argument", _arg_string] = @benchmarkable acos($(nextfloat(zero(T))))
```
Sacha0: `prevfloat`?

pkofod: :+1:

pkofod: Or is that -0.0?

Sacha0: `prevfloat` seems to do the trick? :)
```julia
julia> nextfloat(zero(Float64))
5.0e-324

julia> prevfloat(zero(Float64))
-5.0e-324
```

pkofod:
If it looks good to you, I don't have anything further I want to add.

jrevels:
These will probably have to have their noise tolerances raised later, but might as well wait until there's a need for it before we do so.

pkofod:
> These will probably have to have their noise tolerances raised later, but might as well wait until there's a need for it before we do so.

That's probably true, probably also for the other benchmarks I added! Is it something I should do during the PR-phase?

jrevels:
> That's probably true, probably also for the other benchmarks I added! Is it something I should do during the PR-phase?

Nope, you're good. It's better to adjust the noise tolerances only if they end up being a problem in practice. Otherwise, we'd be decreasing Nanosoldier's ability to catch regressions for no reason.

tkelman:
The `|` are messing with the markdown formatting again

PR #92, opened 2017-07-26
add chaos game fractal benchmark problem

KristofferC:
Creates funny images like:

image

Translated from https://github.com/python/performance/blob/master/performance/benchmarks/bm_chaos.py.

### DIFF
--- a/src/problem/ChaosGame.jl
+++ b/src/problem/ChaosGame.jl
@@ -15,7 +15,7 @@ const DEFAULT_RNG_SEED = 1234
 # GVector #
 ###########
 
-immutable GVector{T <: AbstractFloat}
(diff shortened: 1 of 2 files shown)


src/problem/ChaosGame.jl
```diff
@@ -0,0 +1,267 @@
...
+        end
+    end
+    return s.d[1]
+end
+
+#############
+# ChaosGame #
+#############
+
+immutable ChaosGame{T <: AbstractFloat}
```
StefanKarpinski: Could probably just be Real, no? I'm assuming this doesn't require a floating-point representation.

KristofferC:
Good to go?

PR #93, opened 2017-08-03
update deprecated Bidiagonal constructor

fredrikekre:


jrevels:
Thanks! Can we make this change in a way that's backwards compatible with previous Julia versions?

fredrikekre:
Thanks @ararslan, forgot about this PR :)

PR #94, opened 2017-08-04
Offer master to the Nanosoldier

ararslan:
He hungers for commits


jrevels:
Did the parameters get retuned here, i.e. do the new benchmarks have the correct evaluations-per-sample setting? (sorry I was late on the review)

ararslan:
Not yet. The retuning job should be done on Nanosoldier now so I'll copy it over and submit a PR for that.

jrevels:
Cool beans. For future reference, any new benchmarks merged into the `nanosoldier` branch should be paired with updated parameters in the same merge - otherwise those benchmarks are essentially "wrong" and could result in Nanosoldier reporting spurious timings.

ararslan:
Oh good to know, I didn't realize that. Sorry!!

PR #95, opened 2017-08-05
Retune those benchmarks

ararslan:


ararslan:
I guess there isn't actually anything for you to review here, Jarrett, so I guess I'll just merge.

PR #97, opened 2017-08-06
Change |x| to abs(x)

pkofod:


pkofod:
https://gist.github.com/pkofod/13076ea3e55775fea4e75880bf025a5b is how I **should** look :) (and will do after this)

PR #99, opened 2017-08-17
Add isbits Union benchmarks

quinnj:
I'm assuming this is how I do this right?

### DIFF
--- a/src/misc/MiscellaneousBenchmarks.jl
+++ b/src/misc/MiscellaneousBenchmarks.jl
@@ -9,6 +9,53 @@ using Compat
 const SUITE = BenchmarkGroup()
 
 ###########################################################################
+# isbits Union optimizations for type fields & array elements (issue #22441)
+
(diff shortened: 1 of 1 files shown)


jrevels:
LGTM! Seems like we'll either need to drop <v0.6 support or use the old `type` keyword here to get tests passing.

quinnj:
Drop <v0.6! Long live 0.7!

src/misc/MiscellaneousBenchmarks.jl
```diff
@@ -9,6 +9,53 @@ using Compat
...
+    u::Union{Int64, Void}
+end
+
+function getsetfield(A)
+    for i = 1:length(A)-1
+        A[i].u = A[i + 1].u
+    end
+end
+
+function makeUnionFieldArray(N)
```
ararslan: Would be nice not to use camel case

nalimilan:
Note that we now have a "union" file where these tests should probably go (if they still make sense).

PR #100, opened 2017-08-19
Fix readstring depwarns

yuyichao:


src/io/IOBenchmarks.jl
```diff
@@ -28,7 +28,7 @@ g = addgroup!(SUITE, "read", ["buffer", "stream", "string"])
 testbuf = IOBuffer(randstring(RandUtils.SEED, 10^4))
 
 g["read"]       = @benchmarkable perf_read!($testbuf)
-g["readstring"] = @benchmarkable readstring($testbuf)
+g["readstring"] = @benchmarkable read($testbuf, String)
```
yuyichao: Also, this benchmark seems to be reading empty strings most of the time....

PR #101, opened 2017-08-19
Use propagate_inbounds instead of inline

yuyichao:
Currently the benchmarks are relying on inlining bugs to propagate boundscheck elision
but the bug will be fixed with linear IR.

Ref https://github.com/JuliaLang/julia/issues/18261
This is affecting https://github.com/JuliaLang/julia/pull/23240


yuyichao:
Note that this also raises the version requirement to 0.5. No 0.4 Compat usage have been removed yet and that can be done in a separate PR after either this PR or https://github.com/JuliaCI/BaseBenchmarks.jl/pull/100 is merged.

ararslan:
I'll retune Nanosoldier once this and #100 are merged.

PR #118, opened 2017-09-13
Add benchmark for limited print_matrix (Julia PR 23681)

iamed2:
See https://github.com/JuliaLang/julia/pull/23681


iamed2:
This mainly tests for possible `O(n)` behaviour when we should be seeing `O(1)` due to the fixed display limit.

src/io/IOBenchmarks.jl
```diff
@@ -54,4 +54,19 @@ testdata_buf = serialized_buf(testdata)
...
 
+###################################
+# limited array printing (#23681) #
+###################################
+
+g = addgroup!(SUITE, "array_limit", ["array", "display"])
+
+test_vector = rand(10^8)
+test_column_matrix = reshape(test_vector, length(test_vector), 1)
+test_square_matrix = reshape(test_vector, 10^4, 10^4)
```
stevengj: Aren't these sizes a bit big?  Normally the benchmarks cases should be smaller.  (I think the guideline was for them to run in about 0.01 sec?  I forget.)

PR #119, opened 2017-09-19
Add readuntil benchmarks

omus:
See https://github.com/JuliaLang/julia/pull/20621

### DIFF
--- a/src/string/StringBenchmarks.jl
+++ b/src/string/StringBenchmarks.jl
@@ -29,4 +29,27 @@ g = addgroup!(SUITE, "searchindex")
 g["Char"] = @benchmarkable searchindex($str,  $('x'))
 g["String"] = @benchmarkable searchindex($str, $("x"))
 
+######################
+# readuntil (#20621) #
+######################
+
+buffer = IOBuffer("A" ^ 50000)
+
+g = addgroup!(SUITE, "readuntil")
+for len in (1, 2, 1000, 50000)
+    g["target length $len"] = @benchmarkable readuntil(seekstart($buffer), $("A" ^ len))
+end
+
+buffer = IOBuffer(("A" ^ 50000) * "B")
+target = ("A" ^ 5000) * "Z"
+g["backtracking"] = @benchmarkable readuntil(seekstart($buffer), $target)
+
+buffer = IOBuffer(randstring('A':'X', 40000) * target)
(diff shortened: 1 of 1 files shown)


src/string/StringBenchmarks.jl
```diff
@@ -29,4 +29,27 @@ g = addgroup!(SUITE, "searchindex")
...
+target = ("A" ^ 5000) * "Z"
+g["backtracking"] = @benchmarkable readuntil(seekstart($buffer), $target)
+
+buffer = IOBuffer(randstring('A':'X', 40000) * target)
+target = "Y" * ("Z" ^ 999)
+g["no backtracking"] = @benchmarkable readuntil(seekstart($buffer), $target)
+
+buffer = IOBuffer(("bar" ^ 20000) * "ians")
+target = ("bar" ^ 300) * "ian"
+g["barbarian backtrack"] = @benchmarkable readuntil(seekstart($buffer), $target)
```
ararslan: Good name

omus: In my quest for good backtracking words to use I also found:
* assassin
* cincinatti
* conconscious
* murmurously
* reprepare

ararslan: Wow, all good names.

src/string/StringBenchmarks.jl
```diff
@@ -29,4 +29,27 @@ g = addgroup!(SUITE, "searchindex")
...
+g = addgroup!(SUITE, "readuntil")
+for len in (1, 2, 1000, 50000)
+    g["target length $len"] = @benchmarkable readuntil(seekstart($buffer), $("A" ^ len))
+end
+
+buffer = IOBuffer(("A" ^ 50000) * "B")
+target = ("A" ^ 5000) * "Z"
+g["backtracking"] = @benchmarkable readuntil(seekstart($buffer), $target)
+
+buffer = IOBuffer(randstring('A':'X', 40000) * target)
```
vtjnash: use `randstring(RandUtils.SEED, 'A':'X', 40000)` for stable random values

vtjnash: actually, look like it should be `String(rand(RandUtils.SEED, 'A':'X', 40000))`, so it works on v0.5 and v0.6 too?

PR #123, opened 2017-09-29
atan2 benchmarks

pkofod:
These benchmarks hit all branches of atan2 (unless you're in the case where atan is called, atan benchmarks take care of that). Is this over the top, or good?


pkofod:
relevant for https://github.com/JuliaLang/julia/pull/23468 btw (would be great if nanosoldier was run there after this is merged and tuned)

PR #128, opened 2017-10-06
mixed sparse-dense matrix multiplication benchmarks

Sacha0:
This pull request adds benchmarks for mixed sparse-dense matrix multiplication (in anticipation of improved kernels for these operations).

@pevnak, could you confirm that these benchmarks cover all cases relevant to your application?

These benchmarks might be a bit heavy. If so, is there some flag behind which we can hide some of these benchmarks? If not, perhaps changing the `1` in `    m, k, n = 1 .* (om, ok, on)` to `3` would more realistically exercise cache. (Edit: These benchmarks should become _much_ lighter once improved kernels go live.)

Thanks and best!

### DIFF
--- a/src/sparse/SparseBenchmarks.jl
+++ b/src/sparse/SparseBenchmarks.jl
@@ -157,4 +157,116 @@ for s in sizes
     g["SymTridiagonal", s] = @benchmarkable sparse($(SymTridiagonal(randvec(s), randvec(s-1))))
 end
 
+#########################
+# matrix multiplication #
+#########################
+
+g = addgroup!(SUITE, "matmul")
+
+# mixed sparse-dense matmul #
+#---------------------------#
+
+using Base.LinAlg: *, A_mul_B!,
+    A_mul_Bt,  A_mul_Bc,  A_mul_Bt!,  A_mul_Bc!,
+    At_mul_B,  Ac_mul_B,  At_mul_B!,  Ac_mul_B!,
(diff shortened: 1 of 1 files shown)


pevnak:
This seems to be right to me. The sizes of my matrices are about two magnitudes bigger and even more sparser. I cannot judge, what impact it is going to have on the benchmark though. 

Thanks a lot for this initiative.

Sacha0:
The continuous integration results suggest that these benchmarks took order of a minute, so massaging these benchmarks a bit larger may be alright (@ararslan?).

Also, the benchmarks for adjoint matmul variants will exercise essentially the same kernels as the benchmarks for corresponding transpose matmul variants. (Consider, for example, `A_mul_Bc[!]` and `A_mul_Bt[!]`). The same holds for out-of-place versus in-place matmul variants. So we might save some time by reducing the size of, for example, the adjoint and out-of-place variants, and use that time to increase the size of the transpose and in-place variants.

(That said, I imagine that these benchmarks should be fairly representative of corresponding larger problems as long as they spill out of L2 cache. Increasing `s` to two or three should achieve that.)

Thoughts? Particular suggestions? Thanks and best! :)

ararslan:
> massaging these benchmarks a bit larger may be alright (@ararslan?).

¯\\\_(ツ)\_/¯ Sure, seems reasonable

Sacha0:
I added some scale variation to the matrices in accord with the notes above. Assuming these revised benchmarks run in a reasonable amount of time, this pull request should be in shape. Best!

Sacha0:
Thanks Alex and Tomas! :)

Issue #124, opened 2017-09-30
Benchmarking with interpolated type gives different results


rfourquet:
Sorry for the unclear title. The problem can be summarized as follows: `T = UInt; run(tune!(@benchmarkable rand($T)))` gives a very over-estimated time compared to `run(tune!(@benchmarkable rand(UInt))`. While preparing a PR against julia/master, the RandomBenchmarks showed a lot of regressions because of this (in this case, `T` is set in a loop), even though the performance is not degraded when running individual benchmarks using the second form (i.e. using `UInt` directly). I tried solving this by using some incantation of `eval`, with no success. ~~My last try was something like `T=UInt; RD=RandomDevice(); g[...] = eval(@benchmarkable rand(Expr(:$, RD), $T)` (here `RD` must not be interpolated by `eval`, only by `@benchmarkable`). I'm not sure whether this works as intended~~(edit: it doesn't)~~, but it's ugly, so~~ wanted to discuss this problem here before working more on this.

jrevels:
This behavior is intentional, but it seems to be too poorly documented in BenchmarkTools, because a lot of people run into this. 

In the former example, you're asking for the runtime of a function `f(UInt)` where `@noinline f(T) = rand(T)`. In the second example, you're asking for the performance of `f()` where `@noinline f() = rand(UInt)`. In other words, these benchmarks are measuring different things. The question becomes: What are you interested in measuring?

For example, the former might be considerably slower due to the compiler's choice of type argument specialization, i.e. `Datatype` vs. `Type{T}` (that may or may not be what's going on in your case).

rfourquet:
Thanks for your answer. I understand the problem now, but was not aware of it when writing the random benchmarks. I had intended to measure the the second form (`f() = rand(UInt)`), which correspond to the performance which can be obtained by careful coding (i.e. using e.g. `f(::Type{T})` instead of `f(T::DataType)`). I think I have got a workaround to the difficulty I mentioned (of writing `@eval` around `@benchmarkable` with `$`) using a simple wrapper function. I will submit a PR soon for that.

PR #129, opened 2017-10-08
random: don't benchmark functions with interpolated types

rfourquet:
This fixes part of #124. E.g. When wanting to benchmark
`rand(Int)` and `rand(UInt)`, the following loop won't
measure what is expected:
`for T=(Int, UInt); @benchmarkable rand($T); end`
In order to benchmark the equivalent of
`@benchmarkable rand(Int); @benchmarkable rand(UInt)`,
we "value-fy" the types with `Val`.

### DIFF
--- a/src/random/RandomBenchmarks.jl
+++ b/src/random/RandomBenchmarks.jl
@@ -62,6 +62,13 @@ set_tolerance!(g)
 
 g = addgroup!(SUITE, "types", ["rand", "rand!", "randn", "randn!", "randexp", "randexp!"])
 
+# these functions are used to interpolate values instead of Types in @benchmarkable,
+# to measure the best possible performance, cf. #124
+@inline _rand(        ::Val{T}) where {T} = rand(Base.GLOBAL_RNG, T)
+@inline _rand(   rng, ::Val{T}) where {T} = rand(rng, T)
+@inline _randn(  rng, ::Val{T}) where {T} = randn(rng, T)
+@inline _randexp(rng, ::Val{T}) where {T} = randexp(rng, T)
+
 g["rand",    "ImplicitRNG",     "ImplicitFloat64"] = @benchmarkable rand()
 g["rand",    "RandomDevice",    "ImplicitFloat64"] = @benchmarkable rand($RD)
 g["rand",    "MersenneTwister", "ImplicitFloat64"] = @benchmarkable rand($MT)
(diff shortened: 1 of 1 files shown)


rfourquet:
Gentle bump.

rfourquet:
cc. @jrevels @ararslan

src/random/RandomBenchmarks.jl
```diff
@@ -56,6 +56,13 @@ set_tolerance!(g)
 
 g = addgroup!(SUITE, "types", ["rand", "rand!", "randn", "randn!", "randexp", "randexp!"])
 
+# these functions are used to interpolate values instead of Types in @benchmarkable,
+# to measure the best possible performance, cf. #124
+@inline _rand{T}(::Val{T}) = rand(Base.GLOBAL_RNG, T)
+@inline _rand{T}(rng, ::Val{T}) = rand(rng, T)
+@inline _randn{T}(rng, ::Val{T}) = randn(rng, T)
+@inline _randexp{T}(rng, ::Val{T}) = randexp(rng, T)
```
jrevels: Could we switch this to use `where` syntax?

Other than that, this all LGTM :)

rfourquet: Thanks!! If I remember correctly, I didn't use `where` to satisfy CI (for julia 0.5 I think). I will assume that it's not necessary anymore and will change that tomorrow (it's too late here!).

cstjean:
This LGTM, but I'll suggest that instead of introducing the extra `_rand` functions, you could just use `@eval` on each benchmark, which will splice in the type. To take the example from #124, this:

```julia
T = UInt; @eval run(tune!(@benchmarkable rand($T)))
```

has the same runtime as `run(tune!(@benchmarkable rand(UInt))` on my system.

cstjean:
That is - if I've understood everything correctly 🤔

rfourquet:
I think i tried unsucessfully every combination of `@eval`, `@benchmarkable` and `@quote`... I guess the tricky part is that in the file you don't invoke directly `run (tune!(...))`. But i may have missed something obvious i will recheck.

cstjean:
`g["rand",     "ImplicitRNG",  tstr] = @eval @benchmarkable rand($T)`?

rfourquet:
So you seem to be right! maybe I had tried to put the eval out the loops, or maybe I dismissed this solution because it evals also the rng argument in `@eval @benchmarkable rand($MT, $T)`, which means I think that it (`MT`) is not seen by `@benchmarkable` as a "dollar-value" (the dollar is consumed by `@eval`), but apparently it doensn't make a difference in this case (`MT` is a `const`). Shall I switch to the `@eval` version @jrevels ?

jrevels:
> Shall I switch to the @eval version @jrevels ?

Up to you. The `@eval` trick is nifty, but I think the `_rand` approach is far clearer to read.  It can get pretty hairy to understand what code is doing when splicing is used across several layers of nested macros/evaluation scopes 😛

rfourquet:
> Up to you

Then I will keep it as is! I indeed am uncomfortable with nesting two macros for which `$` has a special meaning. Thanks!

PR #132, opened 2017-10-30
Update for JSON serialization of benchmarks

### LINKED ISSUE #131: Update to use JSON-based serialization
BenchmarkTools now serializes benchmark data to and from JSON rather than JLD, so the tuning here will need to produce a .json file. In all likelihood we will lose the ability to easily retune a specific group only and instead will have to retune everything every time (which is what I've been doing anyway).

ararslan:
The future is here

Fixes #131


README.md
```diff
@@ -82,19 +82,19 @@ following code to execute all benchmarks and save the results (replacing `filena
 an actual unique file name):
 
 ```julia
-using BenchmarkTools, BaseBenchmarks, JLD
+using BenchmarkTools, BaseBenchmarks, JSON
```
jrevels: Do users actually need to bring `JSON` into scope to use `BenchmarkTools.save`/`BenchmarkTools.load`? IIRC the `using JLD` here should've been removed a while ago, I must've just forgotten to do so.

REQUIRE
```diff
@@ -1,4 +1,4 @@
-julia 0.5
-BenchmarkTools 0.0.7
+julia 0.6
+BenchmarkTools 0.2.0
 Compat 0.31.0
-JLD
+JSON
```
jrevels: Similarly, we might not even need JSON here

src/BaseBenchmarks.jl
```diff
@@ -1,17 +1,15 @@
 module BaseBenchmarks
 
 using BenchmarkTools
-using JLD
+using JSON
```
jrevels: and here as well?

ararslan:
You're right, I don't think JSON actually needs to be used or even mentioned directly anywhere in here.

ararslan:
Looks like there's an issue in BenchmarkTools with deserializing a `BenchmarkGroup` full of `BenchmarkGroup`s, which is causing `MethodError`s in `loadall!` here.

README.md
```diff
@@ -82,19 +82,19 @@ following code to execute all benchmarks and save the results (replacing `filena
...
+Next, you can load the results and check for regressions (once again replacing the JSON file
 names used here with the actual file names):
 
 ```julia
-using BenchmarkTools, BaseBenchmarks, JLD
-master = load("master.jld", "results")
-pr = load("pr.jld", "results")
+using BenchmarkTools, BaseBenchmarks
+master = BenchmarkTools.load("master.json")
+pr = BenchmarkTools.load("pr.json")
```
rfourquet: While updating this workflow to this new JSON backend, I noticed that I have to extract the first element of the result of the load for the next lines to work, i.e. `master = BenchmarkTools.load("master.json")[1]`. I don't have time now to investigate if this is always true, but just wanted to mention it somewhere (if it's a simple matter of adding `[1]` in those 2 lines above, I could even open a PR!)

ararslan: The only way to deserialize multiple values in a type-stable way is to put them in an array, so `load` always returns an array. If you've only serialized a single thing, e.g. a single `BenchmarkGroup`, you'll still get an array, so the `[1]` is required in that case. I think in this case you're correct, those lines should have `[1]`. A PR would be welcome!

PR #134, opened 2017-11-04
Fix more 0.7 deprecations

ararslan:


src/problem/SparseFEM.jl
```diff
@@ -1,5 +1,13 @@
 module SparseFEM
 
+using Compat
+
+if VERSION >= v"0.7.0-DEV.2116"
+    _spdiagm(xs, ds) = sparse(SparseArrays.spdiagm_internal(zip(ds, xs)...))
+else
+    _spdiagm = spdiagm
+end
```
ararslan: cc @fredrikekre for a review of this piece in particular

fredrikekre: Presumably you need to use `_spdiagm` somewhere? :)

ararslan: image

src/problem/SparseFEM.jl
```diff
@@ -1,9 +1,17 @@
...
+
 # assemble the finite-difference laplacian
 function fdlaplacian(N)
     # create a 1D laplacian and a sparse identity
-    fdl1 = spdiagm((ones(N-1),-2*ones(N),ones(N-1)), [-1,0,1])
+    fdl1 = _spdiagm((ones(N-1),-2*ones(N),ones(N-1)), [-1,0,1])
```
fredrikekre: I would use the new syntax with `Pair` here instead:
```
_spdiagm(-1 => fill(1., N-1), 0 => fill(-2., N), 1 => fill(1., N-1))
```
Also changed to `fill`, since I know you are quite fond of that transition Alex :trollface: .

src/problem/SparseFEM.jl
```diff
@@ -1,9 +1,17 @@
 module SparseFEM
 
+using Compat
+
+if VERSION >= v"0.7.0-DEV.2116"
```
fredrikekre: ... and update this to
```julia
if VERSION >= v"0.7.0-DEV.2116"
    const _spdiagm = spdiagm # is this const needed?
else
    _spdiagm(a, b, c) = spdiagm((a.second, b.second, c.second), (a.first, b.first, c.first))
end
```

PR #136, opened 2017-11-13
seed random matrices for sparse matmul benchmarks

KristofferC:
Should make the sparse matmul benchmarks less noisy cc @Sacha0


ararslan:
This looks good to me, thanks! I'm going to go ahead and merge this and retune the benchmarks, then we can continue to reduce noise if/as needed.

Sacha0:
Thanks for clearing this! :)

PR #138, opened 2017-11-19
Add benchmarks for ==, isequal, any and all

nalimilan:
Good to have in general, but particularly needed before adapting these functions to support three-valued logic with missing values.


nalimilan:
Bump! (This is blocking JuliaLang/julia#24653.)

ararslan:
It looks okay to me. Nanosoldier is having some currently undiagnosed issues at the moment so benchmarking won't be available right away once this is merged and tuned, just FYI.

PR #140, opened 2017-11-23
Fix for Dates moving to stdlib

ararslan:
Currently causing errors in the Dates tests.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,3 @@
 julia 0.6
 BenchmarkTools 0.2.1
-Compat 0.31.0
+Compat 0.37.0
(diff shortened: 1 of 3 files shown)


vchuravy:
According to Travis the fix is not complete yet?
```
WARNING: Base.Dates is deprecated, run `using Dates` instead.
  likely near /home/travis/.julia/v0.7/JSON/src/Writer.jl:51
ERROR: LoadError: LoadError: type Void has no field TimeType
Stacktrace:
```

ararslan:
Good catch. Looks like I'll be submitting an identical fix to JSON. :)

ararslan:
https://github.com/JuliaLang/METADATA.jl/pull/12192

PR #144, opened 2017-11-27
 Port Nullable benchmark to Union{T, Void} 

nalimilan:
These benchmarks would make more sense with `Missing` (once it is merged), but
in practice `Void` should be equivalent.

Remove benchmarks which no longer make sense for  `get()`, or which are covered
via the recent `any()` and `all()` benchmarks.

Needed to run Nanosoldier on https://github.com/JuliaLang/julia/pull/23642. There are mentions of the "nullable" benchmark in params.json, but I'm not sure what to do about them.


ararslan:
> There are mentions of the "nullable" benchmark in params.json, but I'm not sure what to do about them.

Don't worry about that file, it's effectively internal to Nanosoldier. When benchmarks are retuned (which I do whenever a PR here is merged) that file is updated.

ararslan:
I would keep the `Nullable` benchmarks but only enable them for the appropriate version range. That way we can still check `Nullable` performance on the release-0.6 branch. The check as well as all of the `Nullable` stuff here can be removed once we're done making 0.6.x releases.

nalimilan:
> I would keep the Nullable benchmarks but only enable them for the appropriate version range. That way we can still check Nullable performance on the release-0.6 branch. The check as well as all of the Nullable stuff here can be removed once we're done making 0.6.x releases.

Do we really care at this point? If you want, let's wait until you have tagged 0.6.2.

ararslan:
I don't feel too strongly I suppose. I'll do one last Nanosoldier run on my backporting PR for 0.6.2 then this should be good to go.

PR #147, opened 2017-12-06
Add find benchmarks

nalimilan:
Use a separate file since more benchmarks should be added for other
methods, and on various types.

Will be useful to assess the impact of https://github.com/JuliaLang/julia/pull/24774.

### DIFF
--- a/src/array/ArrayBenchmarks.jl
+++ b/src/array/ArrayBenchmarks.jl
@@ -292,7 +292,7 @@ for x in (x_range, x_vec, Int16.(x_range), Float64.(x_range), Float32.(x_range))
 
     g["==", string(typeof(x_vec), " == ", typeof(x))] =
         @benchmarkable $x_vec == $x
-    g["isequal", string(typeof(x_vec), " isequa l", typeof(x))] =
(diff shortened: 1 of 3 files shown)


src/array/ArrayBenchmarks.jl
```diff
@@ -292,7 +292,7 @@ for x in (x_range, x_vec, Int16.(x_range), Float64.(x_range), Float32.(x_range))
 
     g["==", string(typeof(x_vec), " == ", typeof(x))] =
         @benchmarkable $x_vec == $x
-    g["isequal", string(typeof(x_vec), " isequa l", typeof(x))] =
+    g["isequal", string(typeof(x_vec), " isequal ", typeof(x))] =
```
ararslan: Haha nice catch

ararslan:
Seems reasonable to me, but what do I know? Any idea who would be best to review this?

src/array/ArrayBenchmarks.jl
nalimilan: My own mistake...

nalimilan:
I guess @timholy's +1 is already a good validation.

ararslan:
Works for me.

PR #109, opened 2017-08-31
 add benchmarks for some collection operations

rfourquet:
Maybe not everything here makes sense (so cc-ing @JeffBezanson who may have suggestions), I will review again by tomorrow. Also using `setup` with mutable benchmarks is still a bit tricky for me, I'm not sure to have got it right.

PR #150, opened 2017-12-09
add benchmarks for some collection operations

rfourquet:
~DO NOT MERGE~
This is the same as #109, just to check if it can pass CI (it causes problem with tuning).

### DIFF
--- a/src/collection/CollectionBenchmarks.jl
+++ b/src/collection/CollectionBenchmarks.jl
@@ -1,0 +1,280 @@
+module CollectionBenchmarks
+
+using BenchmarkTools
+
+# TODO: Remove once Compat has BitSet
+if !isdefined(Base, :BitSet)
+    const BitSet = IntSet
+end
+
+const SUITE = BenchmarkGroup()
+
+const MT = MersenneTwister(0)
+
+const iterlen = 1000
+const ints = rand(MT, 1:iterlen, iterlen)
+const strings = [randstring(MT, rand(MT, 1:30)) for i in 1:iterlen]
+const anys = (shuffle!(MT, [ints; strings;])[1:iterlen])::Vector{Any}
+const sortedints = sort(ints)
+
+# random element, in the collections iff 2nd parameter is true
+let anyshuffled = shuffle(MT, anys),
+    filter = VERSION >= v"0.5.0" ? Iterators.filter : Base.filter
+    d = Dict((Int,    true ) => first(filter(i->isa(i, Int),    anyshuffled))::Int,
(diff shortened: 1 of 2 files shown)


rfourquet:
So this is what I suspected: CI passes, even if there is an underlying problem which pops up only when re-tuning. So I updated the offending line by setting `evals` to the number of elements in the collection before `pop!`ing values, hoping this will be enough. Removing WIP, as I don't see what to do else to be sure it will work.

rfourquet:
@ararslan would you mind trying with this new version?

ararslan:
I won't have time for a bit, but you can run the same code that Nanosoldier does: it's in this repo, `etc/tune.jl`.

src/collection/CollectionBenchmarks.jl
```diff
@@ -0,0 +1,279 @@
...
+                  (Vector, String) => strings,
+                  (Vector, Any)    => anys,
+                  (Pair, Int)      => map(Pair, ints, ints),
+                  (Pair, String)   => map(Pair, strings, strings),
+                  (Pair, Any)      => map(Pair, anys, anys),
+                  (BitSet, Int)    => BitSet(ints))
+
+# return a collection that can serve to initialize a container C of type T
+initcoll(C, T) = let initmap = Dict(Vector => Vector,
+                                    Dict   => Pair,
```
mauro3: Can `ObjectIDict` (and maybe `WeakKeyDict`) be added here?

rfourquet: Yes, but as it's not parametric like `Dict`, it will be a little bit of work to include, so I won't have time to do that in this PR. If you want to open a PR based on this one, ping me and I will review.

rfourquet:
@ararslan Thanks for the trick with etc/tune.jl ! Inded it needed a last update, but now it passes on my machine. If CI turns green, would you make me the favor to mege and retune Nanosoldier? It would be great to run this benchmark on two PR I hope to see merged before the feature freeze.

ararslan:
Looks like JSON needs an update for 0.7, since Unicode was just moved to the stdlib. Nanosoldier won't work without that, so I'll try to get that in ASAP.

rfourquet:
> so I'll try to get that in ASAP.

That would be great, but really no worries if you have too little time, there is no reason that I put pressure on you!

ararslan:
Awesome, the PR is passing CI again, so just let me know how the tuning script goes.

rfourquet:
Tuning passes on my end!

ararslan:
Excellent, thanks for checking!

rfourquet:
Thanks for all your thankless work! fingers crossed no...

PR #155, opened 2017-12-18
band-aid for ind2sub & sub2ind

rfourquet:
This allows at least to run the tests! `ind2sub` was necessary to change (insufficient deprecation system, or oversight? may be worth to point to the author of corresponding Julia PR). For `sub2ind`, my guess is that the deprecation made the benchmark so slow that it appeared to take forevever.

### DIFF
--- a/src/array/sumindex.jl
+++ b/src/array/sumindex.jl
@@ -171,10 +171,19 @@ function perf_sumvector_view(A)
     return s
 end
 
+
+if VERSION >= v"0.7.0-DEV.3025"
+    _ind2sub(sz, l) = Tuple(CartesianIndices(sz)[l])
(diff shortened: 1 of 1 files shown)


ararslan:
Thanks!!

timholy:
Thanks indeed! Was there an error in the deprecated path?

rfourquet:
Thanks for chimining in! The deprecation works when computing e.g. `X = ind2sub((100,100,10), 10)`, but not when trying to assign the result of this to a triple, e.g. `a, b, c = X`:
```julia
ERROR: iteration is deliberately unsupported for CartesianIndex. Use `I` rather than `I...`, or use `Tuple(I)...`
```

timholy:
Hmm, for the deprecation we should probably strip the `CartesianIndex` wrapper. Thanks @rfourquet for noticing this.

PR #157, opened 2018-01-06
Prepare SparseFEM benchmark to changes to find() on Julia master

nalimilan:
`find` will soon return cartesian indices (https://github.com/JuliaLang/julia/pull/24774), so this benchmark needs to convert them to linear indices. Unfortunately, `LinearIndices` currently does not allow indexing with an array, so we need to use `map`.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,3 @@
 julia 0.6
 BenchmarkTools 0.2.1
-Compat 0.37.0
+Compat 0.44.0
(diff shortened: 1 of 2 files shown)


nalimilan:
Julia 0.6 needs JuliaLang/Compat.jl#446 to work.

ararslan:
Can you bump the Compat version in REQUIRE to 0.44.0?

nalimilan:
@ararslan Thanks! Can you merge? Doesn't look like I have the needed rights.

ararslan:
Yep, I was waiting for CI to finish, but it turns out I just needed to refresh the page. :)

I'll retune the benchmarks then trigger Nanosoldier on your PR.

Issue #158, opened 2018-01-09
Fix deprecations for 0.7


ararslan:
This issue will be a holding ground for various deprecations in 0.7 that are not yet fixed here.

- [x] `A_mul_B` and friends
- [x] `.'` transpose
- [x] `Void`
- [x] `replace` without `Pair`s
- [x] `search`, `searchindex`
- [x] Implicit assignment to global variables
- [x] Parametric method syntax generated from hdindexing
- [x] `Array` constructor without `uninitialized`
- [x] `sub2ind`
- [x] `CartesianRange`
- [x] `parse(::String)`
- [x] `Vector{UInt8}(::String)`
- [x] `find`
- [x] `parse` in a base without a kwarg
- [x] `method_exists`
- [x] `indmin`/`indmax`

ararslan:
@Sacha0 Would you have time to help with the `A_mul_B` stuff here? Is that possible to do in a way that works in both 0.6 and 0.7?

Sacha0:
(Will have a look at `.'` post #162 :).)

ararslan:
@nalimilan Would you be willing to help out with the `search` and `find` stuff here?

nalimilan:
https://github.com/JuliaLang/Compat.jl/pull/484 should help.

ararslan:
Everything on this list has now been checked off!

PR #160, opened 2018-01-11
Fix numerous deprecation warnings

timholy:
Not tested locally on 0.6, so let's allow the tests to tell me whether I've broken anything. But Compat seems to have been very well updated (thanks to all who did that) for these changes.

Fixes many, but not all, of the issues #158. The rest might be better handled by others. However, I believe @ararslan mentioned that nanosoldier has been slower because of the warnings; I bet this will get the worst of the slowdowns (any indexing-related deprecations are usually fatal for performance).

### DIFF
--- a/src/collection/CollectionBenchmarks.jl
+++ b/src/collection/CollectionBenchmarks.jl
@@ -260,6 +260,7 @@ set_tolerance!(g)
 g = addgroup!(SUITE, "optimizations", ["Dict", "Set", "BitSet", "Vector"])
 
 for T in (Nothing, Bool, Int8, UInt16)
+    local v
     v::Vector{T} = T === Nothing ? Vector{Nothing}(100000) :
                                    rand(MT, one(T):typemax(T), 100000)
     tstr = string(T)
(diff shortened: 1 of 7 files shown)


src/collection/CollectionBenchmarks.jl
```diff
@@ -258,15 +258,16 @@ set_tolerance!(g)
 # cf. issue #20903 and PR #21964
 g = addgroup!(SUITE, "optimizations", ["Dict", "Set", "BitSet", "Vector"])
 
-for T in (Void, Bool, Int8, UInt16)
-    v::Vector{T} = T === Void ? Vector{Void}(100000) :
-                                rand(MT, one(T):typemax(T), 100000)
+for T in (Nothing, Bool, Int8, UInt16)
```
ararslan: The use of `Nothing` here will require a `using Compat` in this submodule in order to work on 0.6.

ararslan:
Thanks so much for your help!

timholy:
OK, this seems to have passed. Merge at will.

ararslan:
Thanks again!!!

PR #165, opened 2018-01-14
Fix `find` benchmarks for `nothing` sentinel

timholy:
This should let BaseBenchmarks run successfully on newest master.


ararslan:
Thanks!

PR #168, opened 2018-01-15
Rewrite .' calls to use transpose

ararslan:
Ref #158

@Sacha0 Does this look right?

### DIFF
--- a/src/micro/methods.jl
+++ b/src/micro/methods.jl
@@ -86,6 +86,12 @@ end
 # randmatstat #
 ###############
 
+if VERSION >= v"0.7.0-DEV.3204"
+    _at_mul_b(A, B) = transpose(A)*B
+else
+    _at_mul_b(A, B) = At_mul_B(A, B)
+end
+
 function perf_micro_randmatstat(t)
     n = 5
     v = zeros(t)
@@ -97,8 +103,8 @@ function perf_micro_randmatstat(t)
         d = randn(n,n)
         P = [a b c d]
         Q = [a b; c d]
-        v[i] = trace((P.'*P)^4)
-        w[i] = trace((Q.'*Q)^4)
+        v[i] = trace(_at_mul_b(P,P)^4)
+        w[i] = trace(_at_mul_b(Q,Q)^4)
     end
     return (std(v)/mean(v), std(w)/mean(w))
 end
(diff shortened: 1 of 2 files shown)


src/micro/methods.jl
```diff
@@ -91,8 +91,8 @@ function perf_micro_randmatstat(t)
         d = randn(n,n)
         P = [a b c d]
         Q = [a b; c d]
-        v[i] = trace((P.'*P)^4)
-        w[i] = trace((Q.'*Q)^4)
+        v[i] = trace((transpose(P)*P)^4)
+        w[i] = trace((transpose(Q)*Q)^4)
```
Sacha0: These rewrites are correct under 0.7, but change dispatch under 0.6 (as e.g. `P.'*P` lowers to `At_mul_B`) in case that is of concern :).

src/sparse/SparseBenchmarks.jl
```diff
@@ -102,13 +102,13 @@ for m in (small_sqr, small_rct, large_sqr, large_rct)
...
-    g["transpose!", s] = @benchmarkable transpose!($(m.'), $m)
+    g["transpose!", s] = @benchmarkable transpose!($(transpose(m)), $m)
     if VERSION >= v"0.7.0-DEV.1415"
         g["adjoint", s] = @benchmarkable adjoint($cm)
-        g["adjoint!", s] = @benchmarkable adjoint!($(cm.'), $cm)
+        g["adjoint!", s] = @benchmarkable adjoint!($(transpose(cm)), $cm)
     else
         g["ctranspose", s] = @benchmarkable ctranspose($cm)
-        g["ctranspose!", s] = @benchmarkable ctranspose!($(cm.'), $cm)
+        g["ctranspose!", s] = @benchmarkable ctranspose!($(transpose(cm)), $cm)
```
Sacha0: In these three cases the functionally equivalent rewrite under 0.7 is `copy(transpose(cm))`; happily that rewrite might work under 0.6 in this case, but not sure :).

ararslan:
Okay, things should be consistent between 0.6 and 0.7 now, so I'm going to go ahead and merge this so that I can clear this out of the deprecations. Thanks for your help, Sacha!

PR #171, opened 2018-01-18
RFC/WIP: Added IdDict to collection benchmarks

mauro3:
Adds the new `IdDict` to the collection benchmarks, x-ref: https://github.com/JuliaLang/julia/pull/25210

IdDict is missing the `pop!(iddict)` method, thus a few more guards.  Probably this method should be added to `IdDict`.

TODO: this should probably be made compatible with older Julia versions.

CC: @rfourquet

### DIFF
--- a/src/collection/CollectionBenchmarks.jl
+++ b/src/collection/CollectionBenchmarks.jl
@@ -6,6 +6,13 @@ using Compat
 if VERSION >= v"0.7.0-DEV.3406"
     using Random
 end
+if VERSION < v"0.7.0-DEV.3439"
+    # This maybe needs updating if Compat.jl is updated for IdDict.
+    struct IdDict end
+    const IDD = false
+else
+    const IDD = true
+end
 
 const SUITE = BenchmarkGroup()
 
@@ -25,7 +32,7 @@ let anyshuffled = shuffle(MT, anys),
              (Int,    false) => first(filter(i->!in(i, ints), rand(MT, 1:iterlen) for _ in 1:typemax(Int)))::Int,
              (String, false) => randstring(MT, 40)::String)
     global randelt
-    randelt(C, T, bool) = (T == Any && (T = Int); C === Dict ? (d[T, bool] => d[T, bool]) : d[T, bool])
+    randelt(C, T, bool) = (T == Any && (T = Int); C === Dict || C === IdDict ? (d[T, bool] => d[T, bool]) : d[T, bool])
 end
 
 const coll = Dict((Vector, Int)    => ints,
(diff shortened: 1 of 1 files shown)


ararslan:
BaseBenchmarks needs to support 0.6 until we're certain there won't be any further 0.6 releases. While it's unlikely, it is the current release version, so we can't be sure of that yet. What I would suggest is to add `IdDict` to Compat first, then use the Compat version here.

mauro3:
Thanks for the speedy feedback! 

It's not possible to add the needed features of IdDict to Compat.jl (its type parameters, see comment https://github.com/JuliaCI/BaseBenchmarks.jl/pull/150#discussion_r156671947). My idea was to just not run IdDict tests on 0.6. If that is not an option, would it be ok to let this PR sit until 0.6 is dropped? Or we could close it and I'll try to remember to re-open later.  Let me know what you think is best.

ararslan:
Ah, okay. In that case, I'd be fine with letting this sit until 0.6 is dropped, or just putting all of the relevant stuff here behind an appropriate version guard thereby disabling it on 0.6. I'd be fine either way. :slightly_smiling_face:

mauro3:
Ok, updated with version guards.  Maybe worth waiting for the verdict on https://github.com/JuliaLang/Compat.jl/issues/464.

src/collection/CollectionBenchmarks.jl
```diff
@@ -41,24 +48,27 @@ initcoll(C, T) = let initmap = Dict(Vector => Vector,
                                     Dict   => Pair,
                                     Set    => Vector,
                                     BitSet => Vector)
+    IDD && (initmap[IdDict] = Pair)
```
rfourquet: As `IdDict` is here to stay, why not put directly `IdDict => Pair` in the initialization of `initmap`? As you defined trivially `IdDict` anyway on 0.6, it won't be an error, it will just not be used, and this is less code to update when version 0.6 support is dropped. Or am I missing something?

src/collection/CollectionBenchmarks.jl
```diff
@@ -41,24 +48,27 @@ initcoll(C, T) = let initmap = Dict(Vector => Vector,
...
 
-for C in (Dict, Set),
+for C in ( IDD ? (Dict, IdDict, Set) : (Dict, Set) ),
         T in (Int, String, Any)
     coll[C, T] = C(initcoll(C, T))
 end
 
 @inline newcoll(::Type{C}, ::Type{T}) where {C,T} = C{T}()
 @inline newcoll(::Type{Dict}, ::Type{T}) where {T} = Dict{T,T}()
+IDD && (@inline newcoll(::Type{IdDict}, ::Type{T}) where {T} = IdDict{T,T}())
```
rfourquet: Same remark as above, maybe the guard is unnecessary here? (idem for `askey` below).

mauro3:
Adressed @rfourquet comments.  I think this could be merged after CI passes (If Compat does something about IdDict (JuliaLang/Compat.jl#464), then it's a one-line update.)

PR #172, opened 2018-01-18
Fixes for LinAlg moving to the stdlib

ararslan:


src/sparse/SparseBenchmarks.jl
```diff
@@ -159,13 +166,13 @@ g = addgroup!(SUITE, "matmul")
...
-    using Base.LinAlg: *, mul!
+    using LinearAlgebra: *, mul!
 else
-    using Base.LinAlg: *, A_mul_B!,
+    using LinearAlgebra: *, A_mul_B!,
         A_mul_Bt,  A_mul_Bc,  A_mul_Bt!,  A_mul_Bc!,
         At_mul_B,  Ac_mul_B,  At_mul_B!,  Ac_mul_B!,
         At_mul_Bt, Ac_mul_Bc, At_mul_Bt!, Ac_mul_Bc!
-end    
+end
```
Sacha0: Need an additional version check within each branch and appropriate using statements IIUC? :)

ararslan: Ah yes, good catch!

PR #176, opened 2018-01-31
Add benchmarks for map and broadcast with Union{T, Missing} arrays

nalimilan:
Add benchmarks comparing a simple loop with `map` and `broadcast`. Use `missing` instead of `nothing` since it propagates with standard operations, which allows testing more cases.

These are the benchmarks associated with https://github.com/JuliaLang/julia/pull/25828.

### DIFF
--- a/src/union/UnionBenchmarks.jl
+++ b/src/union/UnionBenchmarks.jl
@@ -8,9 +8,9 @@ using Compat
 
 const SUITE = BenchmarkGroup()
 
-########################
+###########################
 # array Union{T, Nothing} #
-########################
+###########################
 
 g = addgroup!(SUITE, "array")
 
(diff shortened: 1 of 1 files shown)


ararslan:
These will either need to be behind a version guard or we'll need to introduce a dependency on Missings for 0.6 support.

nalimilan:
Ah, indeed. I've reverted the move to `missing` then, we can use custom methods for now.

src/union/UnionBenchmarks.jl
```diff
@@ -60,14 +84,35 @@ for T in (Bool, Int8, Int64, Float32, Float64, BigInt, BigFloat, Complex{Float64
     X2[samerand(VEC_LENGTH) .> .9] = nothing
     Y2[samerand(VEC_LENGTH) .> .9] = nothing
 
-    for A in (X, X2)
-        g["perf_sum", string(typeof(A))] = @benchmarkable perf_sum($A)
-        g["perf_countnothing", string(typeof(A))] = @benchmarkable perf_countnothing($A)
+    for (M, A) in ((false, X), (true, X2))
+        g["perf_sum", T, M] = @benchmarkable perf_sum($A)
```
nalimilan: I wasn't sure how the benchmarks names work: is it OK to use three dimensions and to use values directly rather than strings?

ararslan: I don't really know. @jrevels?

jrevels: Yes, you can use arbitrary tuples of values for benchmark keys

jrevels: the main guideline is to keep them relatively short - long keys are formatted annoyingly in benchmark reports

nalimilan:
Good to go?

PR #183, opened 2018-02-06
Fix search/searchindex deprecations

ararslan:
@nalimilan Does this look right?

### DIFF
--- a/src/string/StringBenchmarks.jl
+++ b/src/string/StringBenchmarks.jl
@@ -19,13 +19,21 @@ SUITE["join"] = @benchmarkable join($str, $str) time_tolerance=0.40
 
 str = "Gf6FPPWevqer3di13haDSzrRrSiThqmV3k02dALLu7OHdYRR0dfrKf4iCMcDvgZBawx"
 
-g = addgroup!(SUITE, "search")
-g["Char"] = @benchmarkable search($str,  $('x'))
-g["String"] = @benchmarkable search($str, $("x"))
+if VERSION < v"0.7.0-DEV.3272"
+    g = addgroup!(SUITE, "search")
+    g["Char"] = @benchmarkable search($str,  $('x'))
+    g["String"] = @benchmarkable search($str, $("x"))
 
-g = addgroup!(SUITE, "searchindex")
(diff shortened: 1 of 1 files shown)


src/string/StringBenchmarks.jl
```diff
@@ -19,13 +19,21 @@ SUITE["join"] = @benchmarkable join($str, $str) time_tolerance=0.40
...
+    g["String"] = @benchmarkable search($str, $("x"))
+
+    g = addgroup!(SUITE, "searchindex")
+    g["Char"] = @benchmarkable searchindex($str,  $('x'))
+    g["String"] = @benchmarkable searchindex($str, $("x"))
+else
+    # The searchindex deprecation target makes the updated searchindex group
+    # redundant with the updated search group, so they've been combined here
+    g = addgroup!(SUITE, "findfirst")
+    g["Char"] = @benchmarkable findfirst(equalto($('x')), $str)
```
nalimilan: I would have thought that with the recent Compat PR this block would also work on 0.6, doesn't it?

ararslan: Yeah I suppose it should. Would it be better then to put only the `search` benchmarks behind a version guard and unconditionally define these for `findfirst`?

nalimilan: I'd just drop `search`, since `findfirst` is implemented in terms of `search` in Compat anyway, so it's covered.

ararslan: I'm not sure what you mean. We still want to benchmark `search` on 0.6.

nalimilan: I mean that Compat implements `findfirst(::EqualTo, ...)` by simply calling `search`, so if you benchmark the former you actually benchmark the latter on 0.6. No need to benchmark it separately.

ararslan: Looks like nothing in Compat calls `searchindex`, so that should stay I guess.

ararslan: I think I'd rather just keep this how it is in this PR. Seems more explicit.

PR #187, opened 2018-02-20
findall no longer supports generators

ararslan:
since generators are not indexable


nalimilan:
Thanks. Silly me, I forgot I had written benchmarks for all weird cases... Honestly I think we'd better remove the code completely, these benchmarks have been added recently and there's no future for them.

ararslan:
I'd prefer to leave them until we're sure there won't be another 0.6 release, since `find` supports generators in 0.6.

nalimilan:
As you like. Anyway Femtocleaner will remove them at some point.

PR #190, opened 2018-02-21
Remove deprecations (and coming deprecations)

mbauman:
When a deprecation appears in the hot loop of a tuned benchmark, BaseBenchmarks basically becomes non-terminating.  So this PR anticipates some of the coming changes in JuliaLang/julia#24368.  I have also taken the opportunity to fix the few other deprecations I ran into -- none of those were in hot loops.

Note that I have significantly changed some of these algorithms in order to better take advantage of dot-fusion.  I think that is in the spirit of those specific benchmarks, but if desired I can happily revert that back to the most-minimal change required to avoid the deprecation.

### DIFF
--- a/src/array/ArrayBenchmarks.jl
+++ b/src/array/ArrayBenchmarks.jl
@@ -47,7 +47,7 @@ include("sumindex.jl")
 A3d = samerand(11,11,11)
 S3d = view(A3d, 1:10, 1:10, 1:10)
 arrays = (makearrays(Int32, σ, σ)..., makearrays(Float32, σ, σ)..., trues(σ, σ), A3d, S3d)
-ranges = (1:10^5, 10^5:-1:1, 1.0:1e5, linspace(1,2,10^4))
+ranges = (1:10^5, 10^5:-1:1, 1.0:1e5, range(1,stop=2,length=10^4))
 arrays_iter = map(x -> (x, string(typeof(x))), arrays)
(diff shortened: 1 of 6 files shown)


src/array/ArrayBenchmarks.jl
```diff
@@ -47,7 +47,7 @@ include("sumindex.jl")
 A3d = samerand(11,11,11)
 S3d = view(A3d, 1:10, 1:10, 1:10)
 arrays = (makearrays(Int32, σ, σ)..., makearrays(Float32, σ, σ)..., trues(σ, σ), A3d, S3d)
-ranges = (1:10^5, 10^5:-1:1, 1.0:1e5, linspace(1,2,10^4))
+ranges = (1:10^5, 10^5:-1:1, 1.0:1e5, range(1,stop=2,length=10^4))
```
ararslan: Is there Compat for this on 0.6? I think we need to keep supporting 0.6 here (unfortunate as that is) until we're certain there won't be another 0.6 release.

mbauman:
Bump!  Would it be possible to merge this and get it up on Nanosoldier?  I'd like to run JuliaLang/julia#24368 through perf tests.

ararslan:
Thanks Matt! (Both for the contribution and the reminder :wink:) I'll retune the benchmarks and run Nanosoldier on your PR once that's ready.

PR #192, opened 2018-02-28
More deprecation fixes for upcoming JuliaLang/julia#24368

mbauman:
Sorry, I missed these.  I'm not quite sure why they're not printing when run as part of the test suite, but they do print when run individually.


mbauman:
Also, tests will fail until https://github.com/JuliaLang/julia/pull/26263 is merged.

ararslan:
Any idea why the tests still aren't passing even after that PR was merged?

mbauman:
Let's wait to see how my competing PR fares.  It deprecates these things in exactly the opposite way.

PR #195, opened 2018-04-04
Revert #190 and anticipate deprecating implicit scalar broadcasting in setindex!

mbauman:
In anticipation for the new direction in https://github.com/JuliaLang/julia/pull/26347, we must revert #190 (which was anticipating the opposite idea) and then fix the new deprecations.

Sorry for the churn here!


ararslan:
Nice branch name!

ararslan:
Is this good to go now?

mbauman:
Yup!

PR #198, opened 2018-04-25
Fix overflow on cumsum! test

mbauman:
Should fix JuliaLang/julia#26898 (but done blind).


JeffBezanson:
Thanks, this is urgently needed. Somebody please merge.

PR #207, opened 2018-06-19
Add more benchmarks for sum with Union{T,Nothing}

nalimilan:
These are the most straightforward ways of writing such loops, so they should be fast.


nalimilan:
Is the 0.7 hang a known problem?

KristofferC:
Good to go?

PR #212, opened 2018-06-26
disable UTF check when creating input data

KristofferC:
Should fix loading the problem dataset taking forever.

Note that this doesn't change any benchmark, just makes creating the input data set faster.


KristofferC:
`loading group "problem"... done (took 3.460583769 seconds)`

KristofferC:
Yeah, I just see this as a CI unbreaker.

PR #215, opened 2018-06-29
Adapt to new `init` keyword for `reduce` family of functions

andyferris:
Should fix deprecations that caused degradations in https://github.com/JuliaLang/julia/pull/27711


KristofferC:
Need to keep this working on 0.6 as well.

andyferris:
Does this look right?

PR #220, opened 2018-07-05
Use bigger vector for Union benchmarks

nalimilan:
This should limit spurious results for sum. It also reflects better typical use cases (data frame columns with missing values in large datasets).


ararslan:
@nalimilan I'll retune the benchmarks now. If you don't hear from me in a few hours, please ping me with a reminder to finish up. :slightly_smiling_face:

nalimilan:
There's no hurry, I just wanted to do that before I forget.

ararslan:
It's ready now so you can run Nanosoldier again wherever and the change will be reflected.

PR #224, opened 2018-08-15
Set VERSION to 0.7

ararslan:
To trigger FemtoCleaner

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,2 @@
-julia 0.6
-BenchmarkTools 0.2.1
-Compat 0.60.0
+julia 0.7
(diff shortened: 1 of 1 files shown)


KristofferC:
I don't think there is a point femtocleaning this since we still want to be able to test vs 0.6 (imo).

ararslan:
That's fair. I was hoping we wouldn't be doing any more 0.6 releases in which case maintaining 0.6 support here would be unnecessary, but you're right, it's probably better to maintain support for it for now.

KristofferC:
I just want to have it to test 1.0 against.

PR #229, opened 2018-09-19
dont run SpecialFunctions benchmarks by default

KristofferC:
Running all the benchmarks for the special functions take a significant time and I am questioning the usefullness of testing this so carefully for every single nanosoldier run.

* Most of them typically directly call a C-function or they are defined in Base but are "isolated" in the sense that they almost only call very basic functions.
* There are a significant number of them that is aimed to test all the branches of the implementation, which is good for testing the implementation itself but not the julia language.
* There are a huge number of them and they are kinda noisy so it pollutes all the Benchmark reports.

This PR makes them not be included in the `ALL` benchmarkgroup but IIUC you should still be able to run them explicitly using the `specialfunction` group.

cc @jrevels, @pkofod, @ararslan


jrevels:
This definitely seems like a good idea!

> This PR makes them not be included in the `ALL` benchmarkgroup but IIUC you should still be able to run them explicitly using the `specialfunction` group.

This unfortunately is not the case, AFAICT. This PR causes the "specificalfunction" group to not be loaded at all if using `loadall` (which is what Nanosoldier uses), so filtering based on a tag predicate could never return that group (since that group was never there in the first place).

It might make more sense to implement this in Nanosoldier.jl (changing the benchmark script here: https://github.com/JuliaCI/Nanosoldier.jl/blob/master/src/jobs/BenchmarkJob.jl#L336) than here.

pkofod:
They're not important for most nanosoldier runs. It's just important to track that they don't regress over some major changes to the language itself.

PR #234, opened 2018-10-24
WIP: Benchmark flatten

mschauer:
I think I do not understand the concept of `addgroup!` correctly, help welcome. In any case, these Benchmarks would be useful with https://github.com/JuliaLang/julia/pull/29786 in mind.

### DIFF
--- a/src/misc/MiscellaneousBenchmarks.jl
+++ b/src/misc/MiscellaneousBenchmarks.jl
@@ -305,6 +305,17 @@ for N in (1,1000), M in 1:4
     g["zip($(join(fill("1:$N", M), ", ")))"] = @benchmarkable collect($X)
 end
 
+###############################################
+# flatten iterator
+
+g = addgroup!(SUITE, "iterators", ["flatten"])
+let X = Base.Iterators.flatten(fill(rand(50), 100))
+    g["sum(flatten(fill(rand(50), 100))))"] = @benchmarkable sum($X)
+end
+let X = Base.Iterators.flatten(collect((i,i+1) for i in 1:1000))
+    g["sum(flatten(collect((i,i+1) for i in 1:1000))"] = @benchmarkable sum($X)
+end
+
 ####################################################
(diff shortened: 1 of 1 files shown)


src/misc/MiscellaneousBenchmarks.jl
```diff
@@ -305,6 +305,17 @@ for N in (1,1000), M in 1:4
     g["zip($(join(fill("1:$N", M), ", ")))"] = @benchmarkable collect($X)
 end
 
+###############################################
+# flatten iterator
+
+g = addgroup!(SUITE, "iterators", ["flatten"])
```
jrevels: This overwrites the other `"iterators"` group defined above, which was likely not your intent.

Instead, it would probably make sense just to add these benchmarks to this existing `"iterators"` group  (e.g. just delete `g = addgroup!(SUITE, "iterators", ["flatten"])`, add `"flatten"` as a tag to the existing group, and remove the comment separator.

KristofferC: Sorry about that.

Would it make sense for `addgroup!` to error in this case?

KristofferC:
@mschauer I reverted this based on the comments by Jaret.

src/misc/MiscellaneousBenchmarks.jl
mschauer: Hehe, I tend to think that would be a good idea ;-)

mschauer:
Thank you for the help, fixed and rebased at #237

src/misc/MiscellaneousBenchmarks.jl
jrevels: > Would it make sense for addgroup! to error in this case?

Yeah, this would probably be more useful than the current behavior, which is for `addgroup!(suite, id, args...)` to be a convenience function implementing `g = BenchmarkGroup(args...); suite[id] = g; g`.

PR #239, opened 2018-11-02
Retune benchmarks

ararslan:


ararslan:
Still hitting https://github.com/JuliaLang/julia/issues/29579.

KristofferC:
We do that in all the other PRs as well though.

PR #242, opened 2018-12-06
Add some benchmarks for sparse solves

andreasnoack:


andreasnoack:
Failure on 0.6 is unrelated.

PR #249, opened 2019-05-06
Fix Array growth benchmark: interpolate `samerand()` instead of calculating inside benchmark loop

NHDaly:
Speeds up the benchmark run by removing unrelated work from benchmark loop. 

Before this commit, it was incorrectly including the `samerand()` calculation as part of the array growth benchmark.

I don't actually know the right way to list the performance improvements, but I did it like this:
```julia
julia> improvements = regressions(judge(minimum(master), minimum(pr))); # Reversed master & pr to print improvements

julia> pairs = leaves(improvements) # an array of (ID, `TrialJudgement`) pairs
3-element Array{Any,1}:
 (Any["(\"push_single!\", 2048)"], TrialJudgement(+738.94% => regression))
 (Any["(\"push_single!\", 256)"], TrialJudgement(+2911.24% => regression))
 (Any["(\"push_single!\", 8)"], TrialJudgement(+7394.20% => regression))
```

### DIFF
--- a/src/array/ArrayBenchmarks.jl
+++ b/src/array/ArrayBenchmarks.jl
@@ -219,7 +219,7 @@ g = addgroup!(SUITE, "growth", ["push!", "append!", "prepend!"])
 
 for s in (8, 256, 2048)
     vs = samerand(s)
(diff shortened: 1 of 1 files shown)


NHDaly:
The only test that failed is `v0.6`. The other tests all passed.

PR #260, opened 2020-09-05
Add benchmarks for `==(::AbstractString, ::AbstractString)`

omus:
String equality got a performance boost in https://github.com/JuliaLang/julia/pull/37192#issuecomment-687390486 and I wanted to make we don't see a regression in the future.

### DIFF
--- a/src/string/StringBenchmarks.jl
+++ b/src/string/StringBenchmarks.jl
@@ -57,4 +57,18 @@ g["repeat str len 16"] = @benchmarkable repeat("repeatmerepeatme", 500)
 g["repeat char 1"] = @benchmarkable repeat(' ', 500)
 g["repeat char 2"] = @benchmarkable repeat('α', 500)
 
+###################################################
+# ==(::AbstractString, ::AbstractString) (#37192) #
+###################################################
+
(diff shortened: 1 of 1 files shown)


omus:
Failures are unrelated

src/string/StringBenchmarks.jl
```diff
@@ -57,4 +57,27 @@ g["repeat str len 16"] = @benchmarkable repeat("repeatmerepeatme", 500)
...
+g["different"] = @benchmarkable $(SubString(str)) == $(reverse(str))
+
+###################################################
+# ==(::AbstractString, ::AbstractString) (#37467) #
+###################################################
+
+str = String('A':'Z') ^ 100
+g = addgroup!(SUITE, "==(::AbstractString, ::AbstractString)")
+
+g["identical"] = @benchmarkable invoke(==, Tuple{AbstractString, AbstractString}, $str, $str)
```
nalimilan, line 78: Just to be sure: calling `invoke` like this has no overhead compared to defining `f(x, y) = invoke(...)` and benchmarking `f`?

nalimilan (approved):
Looks good to me. I just have a question.

PR #273, opened 2020-12-30
Add benchmark for heterogenous tuples/array literals

Seelengrab:
Ref. https://github.com/JuliaLang/julia/pull/39035


src/misc/MiscellaneousBenchmarks.jl
```diff
@@ -398,4 +398,12 @@ perf_g6() = sum([+(a,b,c,d,e,f) for a in 1:4, b in 1:4, c in 1:4, d in 1:4, e in
 
 SUITE["perf highdim generator"] = @benchmarkable perf_g6()
 
+######################################################
+# Performance of heterogenous tuples #39035 #
+######################################################
+
+perf_het_tuple() = for i in 1:100_000; b = [1, 2.0]; end;                                                    
+
+SUITE["perf heterogenous tuple"] = @benchmarkable perf_arr_lit()
```
crstnbr, line 407: Shouldn't it be `@benchmarkable perf_het_tuple()`?

src/misc/MiscellaneousBenchmarks.jl
```diff
@@ -398,4 +398,12 @@ perf_g6() = sum([+(a,b,c,d,e,f) for a in 1:4, b in 1:4, c in 1:4, d in 1:4, e in
 
 SUITE["perf highdim generator"] = @benchmarkable perf_g6()
 
+######################################################
+# Performance of heterogenous tuples #39035 #
+######################################################
+
+perf_het_tuple() = for i in 1:100_000; b = [1, 2.0]; end;
```
giordano, line 405: Is the loop really necessary?

src/misc/MiscellaneousBenchmarks.jl
Seelengrab: Ah, thank you! I changed the name but forgot to change the function call.

src/misc/MiscellaneousBenchmarks.jl
Seelengrab: It's described as a tuple optimization for `copy_to!` in the linked PR:

> Upon further investigation, 
> it turns out that the issue is that we're allocating extra
> boxed objects when pulling the fields out of the literal,
> since we don't have any specialization for copyto! from
> tuples, falling back to the generic copyto! method, which
> is type unstable in the presence of heterogeneous tuples.

As I understand it, the loop is necessary because it's supposed to time the assignment as well as the copy, so just assigning once would lead to 0 time being spent in the benchmarks as it would be elided as a constant.

Seelengrab: Which it should be anyway as nothing is done with `b` in the first place, but our analysis is not yet sophisticated enough for this.

giordano: I'm moderately sure the issue is with the allocation of the array `[1, 2.0]`:
```
julia> @btime [1, 2.0];
  162.121 ns (6 allocations: 224 bytes)

julia> @btime [1.0, 2.0];
  23.068 ns (1 allocation: 96 bytes)

julia> @btime Any[1, 2.0];
  22.395 ns (1 allocation: 96 bytes)
```
The loop just repeats the operation without adding much insights, and the assignment is pretty much useless for the purpose of the benchmark.

Seelengrab: Gotcha!

Seems it's even more pronounced on my machine, albeit with fewer allocations:

```
julia> f() = b = [1, 2.]
f (generic function with 1 method)

julia> @benchmark f()
BenchmarkTools.Trial:
  memory estimate:  176 bytes
  allocs estimate:  4
  --------------
  minimum time:     1.670 μs (0.00% GC)
  median time:      1.840 μs (0.00% GC)
  mean time:        2.905 μs (0.00% GC)
  maximum time:     206.560 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     10

julia> f() = b = [1., 2.]
f (generic function with 1 method)

julia> @benchmark f()
BenchmarkTools.Trial:
  memory estimate:  96 bytes
  allocs estimate:  1
  --------------
  minimum time:     28.040 ns (0.00% GC)
  median time:      33.769 ns (0.00% GC)
  mean time:        65.205 ns (7.30% GC)
  maximum time:     10.101 μs (97.33% GC)
  --------------
  samples:          10000
  evals/sample:     995
```

Keno:
Should probably be in the arrays category also.

Seelengrab:
Yeah, I wasn't quite sure where to put it, since it's kind of arrays but also not really since it's truly about the type of the literals here 🤔 I'll move it.

Keno (approved):

src/misc/MiscellaneousBenchmarks.jl
giordano: > Seems it's even more pronounced on my machine

I was on Julia v1.6, with v1.5.3 I get pretty much the same result as you

PR #276, opened 2021-02-24
use temp files instead of modifying package

vtjnash:


KristofferC (approved):

KristofferC:
I guess we need to transition to GitHub actions here?

PR #291, opened 2022-01-20
add simple allocation benchmarks

vilterp:
cc @vchuravy 

A simple-as-possible benchmark to help us see the impact of https://github.com/JuliaLang/julia/pull/42768, https://github.com/JuliaLang/julia/pull/43868 etc.

Not sure if I should be doing these loops myself, or letting BenchmarkTools do them for me.

Also, I tried doing a similar thing to allocate a struct, but it seemed to be stack allocated. Any way to get it to go on the heap? Also was thinking of using `@timed` and asserting against `gc_stats.poolalloc` to make sure we're actually measuring pool allocations.

### DIFF
--- a/src/alloc/AllocBenchmarks.jl
+++ b/src/alloc/AllocBenchmarks.jl
@@ -1,0 +1,24 @@
+module AllocBenchmarks
+
+using BenchmarkTools
+
+const SUITE = BenchmarkGroup()
+
+function perf_alloc_many_arrays()
+  a = nothing
+  for _ in 1:10000000
+      a = []
+  end
+end
+
+function perf_alloc_many_strings()
+  a = nothing
+  for i in 1:10000000
+      a = "hello $(i)"
+  end
+end
+
+SUITE["arrays"] = @benchmarkable perf_alloc_many_arrays()
+SUITE["strings"] = @benchmarkable perf_alloc_many_strings()
+
+end # module
--- a/src/BaseBenchmarks.jl
+++ b/src/BaseBenchmarks.jl
@@ -10,6 +10,7 @@ BenchmarkTools.DEFAULT_PARAMETERS.memory_tolerance = 0.01
 const PARAMS_PATH = joinpath(dirname(@__FILE__), "..", "etc", "params.json")
 const SUITE = BenchmarkGroup()
 const MODULES = Dict("array" => :ArrayBenchmarks,
+                     "alloc" => :AllocBenchmarks,
                      "broadcast" => :BroadcastBenchmarks,
                      "collection" => :CollectionBenchmarks,
                      "dates" => :DatesBenchmarks,


src/alloc/AllocBenchmarks.jl
```diff
@@ -0,0 +1,24 @@
...
+
+using BenchmarkTools
+
+const SUITE = BenchmarkGroup()
+
+function perf_alloc_many_arrays()
+  a = nothing
+  for _ in 1:10000000
+      a = []
+  end
```
vilterp, line 11: Not sure if I should be doing these loops myself, or letting BenchmarkTools do them for me.

vtjnash:
Can we add a `GC.safepoint()` call and `Core.Intrinsics.atomic_fence()` (if available) also, so that the compiler is not permitted to defer the assignment to the loop footer and then skip the loop

src/alloc/AllocBenchmarks.jl
```diff
@@ -0,0 +1,47 @@
...
+
+const SUITE = BenchmarkGroup()
+
+function perf_alloc_many_arrays()
+    for _ in 1:10000000
+        # global to ensure that this is heap allocated
+        global a = []
+
+        GC.safepoint()
+        Threads.atomic_fence()
```
vilterp, line 13: @vtjnash does this look right? `Core.Intrinsics.atomic_fence()` took an arg and wasn't sure what to pass, so used this.

vtjnash: yeah, that works (and has been around longer)

vilterp: cool thx!

src/alloc/AllocBenchmarks.jl
```diff
@@ -0,0 +1,58 @@
+module AllocBenchmarks
+
+using BenchmarkTools
+
+const SUITE = BenchmarkGroup()
+
+function perf_alloc_many_arrays()
+    for _ in 1:10000000
+        # global to ensure that this is heap allocated
+        global a = []
```
vchuravy, line 10: Maybe easier to escape into a:

```
@noinline escape(x) = (Base.inferencebarrier(x); return nothing)
```

vilterp: Huh, in place of `global a = []`? Not quite sure what the `escape` and `inferencebarrier` are doing here

vtjnash: That is just a pure function that returns `nothing`–we don't need to infer that to remove it

NHDaly: I'm also confused by this. Are you saying this is the suggested diff?:
```suggestion
        escape([])
```
with your function defined elsewhere?

Does that inferencebarrier actually prevent the compiler from compiling away the function call, even though `escape()` ultimately returns `nothing`?

-----------

Also why is this preferable to what @vilterp has already? (🙏 Plz teach me to write good benchmarks 🙏)

vilterp:
here's a gist of the results, running on https://github.com/JuliaLang/julia/pull/43868 vs master. Shows no regressions

https://gist.github.com/vilterp/2444a133822427238e46eb23f5b60925

src/alloc/AllocBenchmarks.jl
NHDaly, line 10: Talking with @vtjnash, we don't think the inferencebarrier will be enough.

We also think that the overhead of the atomic fence here might be too high..

maybe switch to allocating a thousand things per alloc, like something like this:
```suggestion
        global a = [[] for _ in 1:1000]
```

vilterp: done

vilterp:
Moved some work to inner loops to offset the impact of calling `Threads.atomic_fence()`

NHDaly (approved):
This LGTM. Thanks for setting it up! :) 👍 👍

NHDaly:
Excellent, thanks!

PR #292, opened 2022-02-01
add tags to allocs benchmark group

vilterp:
Maybe this is why we haven't been able to run it here? https://github.com/JuliaLang/julia/pull/43868#issuecomment-1025907596

cc @NHDaly


vtjnash (changes requested):
This mapping is handled here https://github.com/JuliaCI/BaseBenchmarks.jl/blob/bcec2143f4f956790b40d12f71dcfa76d15a5460/src/BaseBenchmarks.jl#L13

vilterp:
ok cool.

PR #294, opened 2022-03-02
Add more benchmarks for `missing`

nalimilan:


nalimilan:
Thanks! Do you know who is able to retune Nanosoldier so that the new benchmarks can be used?

vtjnash:
They can be used without tuning

PR #300, opened 2022-09-05
inference: add benchmark target check performance of concrete-eval

aviatesk:


aviatesk:
@Keno do these example targets seem reasonable to you?

aviatesk:
I dropped a benchmark target for semi-concrete interpretation, as I'm still trying to figure out what kind of code is supposed to be improved by the semi-concrete eval.

PR #302, opened 2022-09-30
adjust to JuliaLang/julia#46966

aviatesk:


vtjnash:
This seems to fail on julia 1.8 and later with `UndefVarError: StmtInfo not defined`?
https://github.com/JuliaCI/BaseBenchmarks.jl/actions/runs/3158463166/jobs/5140521906#step:5:7978

aviatesk:
> This seems to fail on julia 1.8 and later with `UndefVarError: StmtInfo not defined`? https://github.com/JuliaCI/BaseBenchmarks.jl/actions/runs/3158463166/jobs/5140521906#step:5:7978

Yeah, I mistakenly made the condition upside down.

PR #325, opened 2024-11-13
InferenceBenchmarks: allow benchmark against `Compiler` as the stdlib

aviatesk:
This is especially useful for optimizing the compiler’s implementation while using Revise in a local environment.

### DIFF
--- a/src/inference/InferenceBenchmarks.jl
+++ b/src/inference/InferenceBenchmarks.jl
@@ -18,7 +18,15 @@ module InferenceBenchmarks
 # managed by the runtime system: this allows us to profile Julia-level inference reliably
 # without being influenced by previous trials or some native execution
 
+@static if VERSION ≥ v"1.12.0-DEV.1581"
+if Base.REFLECTION_COMPILER[] === nothing
+const CC = Base.Compiler
+else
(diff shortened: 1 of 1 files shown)


src/inference/InferenceBenchmarks.jl
```diff
@@ -18,7 +18,15 @@ module InferenceBenchmarks
 # managed by the runtime system: this allows us to profile Julia-level inference reliably
 # without being influenced by previous trials or some native execution
 
+@static if VERSION ≥ v"1.12.0-DEV.1581"
+if Base.REFLECTION_COMPILER[] === nothing
```
Keno, line 22: I'm a little concerned about this one. This is a runtime variable, but used in a compile-time context. It probably works fine for now, but we may want to do something like eval things on the fly when requested.

aviatesk: An `eval`-based implementation seems more correct. We need to be cautious about whether it affects benchmarks though. Currently the benchmark modules in BaseBenchmarks.jl are not precompiled, and since these benchmarks are essentially runtime-only, this approach should be fine for now.

PR #349, opened 2026-07-11
Add a benchmark group for global variable access

Keno:
Add a "globals" group covering the access patterns whose performance depends on how the runtime and codegen treat global bindings: reads, stores, and read-modify-writes of typed globals (including a kernel touching 16 distinct bindings per iteration), swapglobal!, and untyped and const reads for reference. The kernels pair each read with a store so the accesses cannot be hoisted out of the benchmark loop.

For use with https://github.com/JuliaLang/julia/pull/62335


vtjnash (approved):
SGTM. Can you add some for `getglobal(m, compilerbarrier(s))` and `isdefined`?

